PR #26, opened 2019-10-29
Add unit tests for Spectrum M4i

peendebak:
Add simple tests for M4i driver.  (should have been part of #24)

@astafan8

### DIFF
--- a/qcodes_contrib_drivers/tests/test_Spectrum_M4i.py
+++ b/qcodes_contrib_drivers/tests/test_Spectrum_M4i.py
@@ -14,7 +14,7 @@ class TestM2j(unittest.TestCase):
             import qcodes_contrib_drivers.drivers.Spectrum.M4i
             m4i = qcodes_contrib_drivers.drivers.Spectrum.M4i.M4i('test_m4i_instrument')
             self.addCleanup(qcodes_contrib_drivers.drivers.Spectrum.M4i.M4i.close_all)
-    
+
             m4i.sample_rate()
             self.mock_pyspcm_module.int32.assert_called()
             m4i.wait_ready()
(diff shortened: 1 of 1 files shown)


qcodes_contrib_drivers/tests/test_Spectrum_M4i.py
```diff
@@ -0,0 +1,21 @@
...
+
+        with patch.dict('sys.modules', pyspcm=mock_pyspcm_module):
+            import qcodes.instrument_drivers.Spectrum.M4i
+            m4i = qcodes.instrument_drivers.Spectrum.M4i.M4i('test_m4i_instrument')
+
+            m4i.sample_rate()
+            mock_pyspcm_module.int32.assert_called()
+            m4i.wait_ready()
+            mock_pyspcm_module.spcm_dwSetParam_i32.assert_called()
+            m4i.close()
```
astafan8: Please move setting up the mock and instantiating the instrument to setup method and closing the instrument (preferably via its class `M4i.cloase_all()`) to teardown method of the test case class.

Also, I recommend using pytest instead of unittest.

eendebakpt: @astafan8 How can I move the instrument instantiating to the `setUp`? The instrument needs a patch of the `pyspcm` module which I now do with a `with patch.dict(...)`. I would like to keep the patch atomic and not spread it out over a setUp and tearDown.

astafan8: > I would like to keep the patch atomic and not spread it out over a setUp and tearDown

in this case, ok, don't move the instatiation of the instrument to setUp but then right after it's initiated please add an "automatic closing the instrument" using [`self.addCleanup` method](https://docs.python.org/3/library/unittest.html#unittest.TestCase.addCleanup) (for example `self.addCleanup(M4i.close_all)`)

PR #32, opened 2020-01-08
Andor CCD DU401 driver

lankes-fzj:
New driver for 'Andor CCD DU401'

(based on SvenBo90/Qcodes@2cff13759589cabf765f744afb448fdec658e73a)


astafan8:
@lankes-fzj see commit https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/32/commits/e8344e05ee7ee7e8d66c461ca93e7f803d876c50 about `WinDLL` type annotation and `sys.platform` usage. Otherwise type checking with `mypy` fails the CI.

lankes-fzj:
Do you have any idea, why mypy fails with the error "Cannot determine type of 'dll'"? This didn't happen to PRs #28 and #29...

jenshnielsen:
I am not sure but I suggest that you just put an `assert sys.platform == 'win32'` at the top of the file just after the imports as suggested here https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks

lankes-fzj:
> I am not sure but I suggest that you just put an `assert sys.platform == 'win32'` at the top of the file just after the imports as suggested here https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks

But then the tests would fail on linux, wouldn't they?

jenshnielsen:
Which tests? As documented in the link above this will make mypy stop doing any cheking of that file on platforms that are not windows. As far as I can see there are no other tests for this at the moment

jenshnielsen:
Sorry that fails the docs build as it cannot import the code :( 

The best workaround that I can come up with is to use cdll on linux.

I think the following will work (even if its a bit strange) 

```python
    def __init__(self, dll_path: Optional[str] = None, verbose: bool = False):
        if sys.platform != 'win32':
            self.dll = ctypes.cdll.LoadLibrary(dll_path or self._dll_path)
            raise OSError("\"atmcd64d\" is only compatible with Microsoft Windows")
        else:
            self.verbose = verbose
            self.dll = ctypes.windll.LoadLibrary(dll_path or self._dll_path)
```

lankes-fzj:
@jenshnielsen Thank you very much for your help :)

PR #33, opened 2020-01-08
Standa 10MWA168 driver

lankes-fzj:
New driver for 'Standa 10MWA168' filter wheel

(based on SvenBo90/Qcodes@60e4cf853fd7cd50937a2e0ec77dd22801ce55e4)


qcodes_contrib_drivers/drivers/Standa/Standa_10MWA168.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def __init__(self, dll_path=None, verbose=False):
+        # save attributes
+        self.verbose = verbose
+
+        # connect to the dll
+        current_path = os.getcwd()
+        os.chdir(os.path.dirname(self._dll_path))
+        self.dll = ctypes.windll.LoadLibrary(dll_path or self._dll_path)
+        os.chdir(current_path)
```
astafan8: plase wrap this in `try-finally` as here https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/28

qcodes_contrib_drivers/drivers/Standa/Standa_10MWA168.py
```diff
@@ -0,0 +1,206 @@
...
+                               3.0e-8: 29, 3.0e-9: 30})
+
+        self.add_parameter('position',
+                           set_cmd=self.set_position,
+                           get_cmd=self.get_position,
+                           get_parser=float,
+                           label='Position')
+
+        self.add_parameter('status',
+                           get_cmd=self.get_status,
```
astafan8: if parameters are supposed to be used, and not the `set_*` methods explicitly, then please make the `set_` methods private (with underscore in the front).

qcodes_contrib_drivers/drivers/Standa/Standa_10MWA168.py
```diff
@@ -0,0 +1,206 @@
...
+
+        # set position of the second wheel
+        self.position.set(int(position_wheel_2))
+        time.sleep(1)
+        for i in range(100):
+            if self.status.get() == 0:
+                break
+
+        # wait another time
+        time.sleep(10)
```
astafan8: it's a good practice to implement these sleep periods via attributes (private or public, or even as properties) so that they can be changed at runtime without modifying the code of the driver.

qcodes_contrib_drivers/drivers/Standa/Standa_10MWA168.py
```diff
@@ -0,0 +1,206 @@
...
+                           np.ceil(current_position / self.revolution + 2) * self.revolution
+        position_wheel_1 = self.offset_wheel_1 + self.distance * filter_wheel_1 + \
+                           np.ceil(current_position / self.revolution + 2) * self.revolution
+
+        if position_wheel_1 > position_wheel_2:
+            print('new', position_wheel_1, position_wheel_2)
+            position_wheel_1 -= self.revolution
+
+        # set position of the second wheel
+        self.position.set(int(position_wheel_2))
```
astafan8: perhaps `np.floor` or python's `floor` functions were meant to be used instead of `int`? those express the purpose in a much cleaner way (if the purpose is to round down)

jenshnielsen:
I think you need to mark the notebook as non executable as documented here https://qcodes.github.io/Qcodes/examples/writing_drivers/Creating-Instrument-Drivers.html#Documentation

We are trying to execute the notebooks in ci but that cannot work as the ci server does not have the instrument installed

lankes-fzj:
> please fix the example notebook to use the correct import (https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=7294&view=logs&j=347123c7-c6ed-5812-0c14-06a064703b24&t=6099742c-3a7b-551c-289a-27a3d6f068ec&l=85). Also, some additionaly info on how to use the instrument would be beneficial to users (but that's optional). Then we can merge.

Added never-execute flag to the notebook's metadata

qcodes_contrib_drivers/drivers/Standa/Standa_10MWA168.py
```diff
@@ -0,0 +1,211 @@
+from qcodes import Instrument
+import ctypes
+import numpy as np
+import os
+import time
+
+
+class DeviceEnumeration(ctypes.BigEndianStructure):
+    pass
```
jenshnielsen: The docs build fails because there are no docs for the `BigEndianStructure` in the main python docs. 

Do you know the purpose of `DeviceEnumeration`? It does not seem to be used anywhere in the code. 
In that case it may be easier to just delete it. 

Otherwise you can add an ignore to `nitpick_ignore` in  `docs/conf.py` 
I think it should be `('py:class', 'ctypes._endian.BigEndianStructure')`

lankes-fzj: Oh yes, it really isn't used anywhere. Since I got the drivers from someone else, I do not know what purpose this class should serve. So I'll simply remove it.

PR #37, opened 2020-02-02
Driver/ni daq

loganbvh:
Driver for National Instruments multifunction DAQ devices (https://www.ni.com/en-us/shop/select/multifunction-io-device). Requires the nidaqmx python package from National Instruments (https://nidaqmx-python.readthedocs.io/en/latest/).


qcodes_contrib_drivers/drivers/NationalInstruments/DAQ.py
```diff
@@ -0,0 +1,166 @@
...
+        # We need parameters in order to write voltages in a qcodes Loop or Measurement
+        for ch, idx in channels.items():
+            self.add_parameter(
+                name='voltage_{}'.format(ch.lower()),
+                dev_name=dev_name,
+                idx=idx,
+                parameter_class=DAQAnalogOutputVoltage,
+                label='Voltage',
+                unit='V'
+            )
```
GateBuilder:
```suggestion
            )
            
```

PR #38, opened 2020-02-02
Driver/attocube

loganbvh:
Qcodes driver for Attocube controllers, like ANC300 (https://www.attocube.com/en/products/nanopositioners/motion-controllers-overview/anc300)

### DIFF
--- a/qcodes_contrib_drivers/drivers/Attocube/AttocubeController.py
+++ b/qcodes_contrib_drivers/drivers/Attocube/AttocubeController.py
@@ -15,7 +15,7 @@ class AttocubeController(VisaInstrument):
     voltage limits to protect the Attocubes.
 
     I have found the Attocube serial communication to be kind of sketchy,
-    so ask_raw() is overridden try to make communication more robust. YMMV.
+    so ask_raw() is overwritten to try to make communication more robust. YMMV.
 
     Args:
         atto_config (dict): Configuration dict (example below).
--- a/qcodes_contrib_drivers/drivers/Attocube/AttocubeController.py
+++ b/qcodes_contrib_drivers/drivers/Attocube/AttocubeController.py
@@ -1,0 +1,324 @@
+from typing import List, Dict, Union, Optional, Sequence, Any, Tuple
+import numpy as np
+import visa
+import time
+import logging
+log = logging.getLogger(__name__)
+
+from qcodes.instrument.visa import VisaInstrument
+import qcodes.utils.validators as vals
+
+class AttocubeController(VisaInstrument):
+    """Base class for Attocube controller instrument.
+
+    This driver "low temperature" and "room temperature" modes, with different
+    voltage limits to protect the Attocubes.
+
+    I have found the Attocube serial communication to be kind of sketchy,
+    so ask_raw() is overridden try to make communication more robust. YMMV.
+
+    Args:
+        atto_config (dict): Configuration dict (example below).
+        temp (str): 'LT' or 'RT', depending on whether the attocubes are at room temp. or low temp.
+        kwargs: Keyword arguments to be passed to VisaInstrument constructor.
+
+    An example of the atto_config dictionary is provided below:
+        {
+            "name": "atto",
+            "model": "ANC300",
+            "address": "ASRL1::INSTR",
+            "timeout": 5,
+            "terminator": "\r\n",
+            "stopbits": 1,
+            "baud_rate" : 38400,
+            "axes": {"x": 1, "y": 2, "z": 3},
+            "voltage_limits": {
+                "RT":{"x": 25,"y": 25,"z": 25}, # V
+                "LT":{"x": 60, "y": 60, "z": 60} # V
+            },
+            "default_frequency": {"x": 100, "y": 100, "z": 100}, # Hz
+            "history": {}
+        }
+    """
+    def __init__(self, atto_config: Dict, temp: str, ureg: Any, str, **kwargs) -> None:
+        super().__init__(atto_config['name'], atto_config['address'], atto_config['timeout'],
+                         atto_config['terminator'], **kwargs)
+        if temp.upper() not in ['LT', 'RT']:
+            raise ValueError('temp must be "LT" or "RT".')
+        self.metadata.update(atto_config)
+        self.visa_handle.baud_rate = atto_config['baud_rate']
+        self.visa_handle.stop_bits = visa.constants.StopBits.one
+        self.visa_handle.parity = visa.constants.Parity.none
+        self.visa_handle.read_termination = '\r\n'
+        _ = self.parameters.pop('IDN') # Get rid of this parameter
+        self.temp = temp
+        self.timestamp_fmt = r'%Y-%m-%d_%H:%M:%S'
+        self.axes = self.metadata['axes']
+        self.voltage_limits = {}
+        self.add_parameter('version',
+                            label='Version',
+                            get_cmd='ver',
+                            get_parser=str,
+                            snapshot_get=False
+                            )
+        for axis, idx in self.axes.items():
+            self.voltage_limits.update({axis: self.metadata['voltage_limits'][temp][axis]}) # Volts
+            # axis mode
+            self.add_parameter('mode_ax{}'.format(idx),
+                                label='{} axis mode'.format(axis),
+                                unit='',
+                                get_cmd='getm {}'.format(idx),
+                                set_cmd='setm {} {{}}'.format(idx),
+                                vals=vals.Enum('gnd', 'inp', 'cap', 'stp', 'off', 'stp+', 'stp-'),
+                                get_parser=self._mode_parser,
+                                snapshot_get=False
+                                )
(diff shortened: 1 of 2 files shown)


qcodes_contrib_drivers/drivers/Attocube/AttocubeController.py
```diff
@@ -0,0 +1,324 @@
...
+import logging
+log = logging.getLogger(__name__)
+
+from qcodes.instrument.visa import VisaInstrument
+import qcodes.utils.validators as vals
+
+class AttocubeController(VisaInstrument):
+    """Base class for Attocube controller instrument.
+
+    This driver "low temperature" and "room temperature" modes, with different
```
GateBuilder:
```suggestion
    This driver has "low temperature" and "room temperature" modes, with different
```

qcodes_contrib_drivers/drivers/Attocube/AttocubeController.py
```diff
@@ -0,0 +1,324 @@
...
+    voltage limits to protect the Attocubes.
+
+    I have found the Attocube serial communication to be kind of sketchy,
+    so ask_raw() is overridden try to make communication more robust. YMMV.
```
GateBuilder:
```suggestion
    so ask_raw() is overwritten to try to make communication more robust. YMMV.
```

qcodes_contrib_drivers/drivers/Attocube/AttocubeController.py
```diff
@@ -0,0 +1,324 @@
...
+
+    Args:
+        atto_config (dict): Configuration dict (example below).
+        temp (str): 'LT' or 'RT', depending on whether the attocubes are at room temp. or low temp.
```
GateBuilder: Could we please remove the types from the docstring? They already defined in the signature.

qcodes_contrib_drivers/drivers/Attocube/AttocubeController.py
```diff
@@ -0,0 +1,324 @@
...
+            'LT': {'x': 60, 'y': 60, 'z': 60} # low temperature limits, V
+        },
+        'default_frequency': {'x': 100, 'y': 100, 'z': 100}, # stepping frequency, Hz
+    }
+    """
+    def __init__(self, atto_config: Dict, temp: str, **kwargs) -> None:
+        super().__init__(atto_config['name'], atto_config['address'], atto_config['timeout'],
+                         atto_config['terminator'], **kwargs)
+        if temp.upper() not in ['LT', 'RT']:
+            raise ValueError('temp must be "LT" or "RT".')
```
GateBuilder:
```suggestion
            raise ValueError('Temperature must be "LT" or "RT".')
```

qcodes_contrib_drivers/drivers/Attocube/AttocubeController.py
```diff
@@ -0,0 +1,324 @@
...
+            cmd: Command to write to controller.
+
+        Returns:
+            (str): Response of Attocube controller to the query `cmd`.
+        """
+        self.device_clear()
+        response = super().ask_raw(cmd)
+        status = super().ask_raw(cmd)
+        self.check_response(status)
+        time.sleep(0.2)
```
GateBuilder: Why do we need to sleep?

qcodes_contrib_drivers/drivers/Attocube/AttocubeController.py
```diff
@@ -0,0 +1,324 @@
...
+        """Performs a given number of Attocube steps. steps > 0 corresponds to stepu (up),
+        steps < 0 corresponds to stepd (down).
+        
+        Args:
+            axis: Either axis label (str, e.g. 'y') or index (int, e.g. 2)
+            steps: Number of steps to perform (>0 for 'u', <0 for 'd')
+        """
+        axis, idx = self._parse_axis(axis)
+        if not isinstance(steps, int):
+            raise ValueError('steps must be an integer.')
```
GateBuilder:
```suggestion
            raise ValueError('Steps must be an integer.')
```

qcodes_contrib_drivers/drivers/Attocube/AttocubeController.py
```diff
@@ -0,0 +1,324 @@
...
+        freq = getattr(self, 'freq_ax{}'.format(idx))()
+        log.info('Performing {} steps along axis {}.'.format(steps, axis))
+        getattr(self, 'mode_ax{}'.format(idx))('stp')
+        direc = 'u' if steps > 0 else 'd'
```
GateBuilder: Can we write this logic flow normally instead of a single liner for readability? As I found this function already quite hard to read. May be some comments along the way would be helpful.

qcodes_contrib_drivers/drivers/Attocube/AttocubeController.py
```diff
@@ -0,0 +1,324 @@
...
+        self.add_parameter('version',
+                            label='Version',
+                            get_cmd='ver',
+                            get_parser=str,
+                            snapshot_get=False
+                            )
+        for axis, idx in self.axes.items():
+            self.voltage_limits.update({axis: self.metadata['voltage_limits'][temp][axis]}) # Volts
+            # axis mode
+            self.add_parameter('mode_ax{}'.format(idx),
```
GateBuilder: Instead of `.format()`, we prefer f-string notation when possible. Here, for example, instead of `"mode_ax{}".format(idx)`, preferred notation would be `f"mode_ax{idx}"`. This comment applies to the rest of the code.

qcodes_contrib_drivers/drivers/Attocube/AttocubeController.py
loganbvh: Good point, done here: https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/38/commits/4832b40ef5d657b3bd72652ed08dd37279402732

qcodes_contrib_drivers/drivers/Attocube/AttocubeController.py
loganbvh: See https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/38/commits/5ed9b8ad9a87e02a13ba02a34d0e04b2a5c8ddcb

qcodes_contrib_drivers/drivers/Attocube/AttocubeController.py
loganbvh: Okay, done here https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/38/commits/9ec46d3f25aab9b62f236963b46b550ac71d2900

qcodes_contrib_drivers/drivers/Attocube/AttocubeController.py
loganbvh: You're right that we might not need to sleep. As noted in the docstring, I had a hard time getting serial communication to work reliably with these instruments when I originally wrote this driver. I know that it works reliably with the sleep. I no longer have access to the instrument itself, so I unfortunately can't test the driver without the sleep.

MichaelWagener:
I've looked into this code. It has a slightly different interface than my code but I can try this at our device and so we can unify both. Should we discuss this here or contact directly via mail?

loganbvh:
@MichaelWagener we can discuss here. It looks like you have implemented a few more features than I have (AC/DC inputs and triggered outputs), and you are using qcodes `InstrumentChannel` and `ChannelList`, which is probably preferable to the way I implemented multiple axes.

My driver has been in use since 2018 on multiple ANC300 devices with three ANM150 modules each (as well as with an older ANC150 device). Unfortunately I don't currently have access to these instruments, so I can't test out your version of the driver. If you would like to try my driver on your devices and suggest changes for a unified driver, that would be great.

One question I have is: why are you using pyserial instead of qcodes' built-in serial communication via pyvisa?

MichaelWagener:
@loganbvh yes, I can test your driver in the next days and will report. And yes, it’s better to switch to pyvisa. I’ll do it.
I would prefer to put the communication parameters in the constructor parameter string and then ask all version numbers of the axis controller. If there was an error, this modul is not installed. And I think then it should be a good idea to check the version numer to get the limits and enable other functions (such as AC/DC and trigger outputs). That is in my mind and not programmed at the moment. I think that would be better than a dict as you do it.

qcodes_contrib_drivers/drivers/Attocube/AttocubeController.py
GateBuilder: Ok, no problem it can stay. I just wanted to know if there is a specific reason.

MichaelWagener:
Just for information: my driver is ready (https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/34) and tested with my device. I switched to pyvisa, the initialization was copied from your code. Thanks for that.

loganbvh:
@GateBuilder It is fine with me if you go ahead and merge @MichaelWagener PR (https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/34) and close mine

PR #39, opened 2020-02-02
Driver/aimtti

loganbvh:
Qcodes driver for Aim-TTi EL320P, the power supply used for Bluefors for their 4 K stage warmup heater.

### DIFF
--- a/qcodes_contrib_drivers/drivers/AimTTi/EL320P.py
+++ b/qcodes_contrib_drivers/drivers/AimTTi/EL320P.py
@@ -76,3 +76,4 @@ class EL320P(VisaInstrument):
 
     def _error_parser(self, response):
         return int(response[-1])
+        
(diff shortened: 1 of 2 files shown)


qcodes_contrib_drivers/drivers/AimTTi/EL320P.py
```diff
@@ -0,0 +1,81 @@
...
+            name='current_out',
+            label='Actual output current',
+            unit='A',
+            get_cmd='IO?',
+            get_parser=self._get_parser
+            )
+        self.add_parameter(
+            name='mode',
+            label='Mode',
+            unit='',
```
GateBuilder: If there is no applicable units, let us just drop `unit`? There are other parameters with `unit` is defined as empty string, my comment applies also to them.

qcodes_contrib_drivers/drivers/AimTTi/EL320P.py
```diff
@@ -0,0 +1,81 @@
...
+        self.connect_message()
+
+    def _get_parser(self, response):
+        return float(response[2:])
+
+    def _output_parser(self, response):
+        return response[4:]
+
+    def _error_parser(self, response):
+        return int(response[-1])
```
GateBuilder: We need a new line here.

PR #48, opened 2020-04-02
Driver for NI RFSG signal generators

mgunyho:
Hello!

This PR contains a driver for signal generators based on the National Instruments RFSG driver. It also has a generic DLL wrapper for NI drivers, and we have another driver based on that that we could publish once this PR has been merged.


qcodes_contrib_drivers/drivers/NationalInstruments/RFSG.py
```diff
@@ -0,0 +1,245 @@
...
+        Args:
+            source: reference clock source. Valid values are
+                        onboard
+                        clk_in
+                        ref_in
+                        pxi_clk
+                        ref_in_2
+                        pxi_clk_master
+                    Last two values are valid for PXIe-5840 with PXIe-5653.
+            clock_rate: Not implemented. Reference clock frequency in Hertz.
```
astafan8, line 204: if this is not implemented, shall it not be a part of the interface for clarity?

qcodes_contrib_drivers/drivers/NationalInstruments/RFSG.py
```diff
@@ -0,0 +1,245 @@
...
+        """
+        if source in CLK_SRC_MAP.keys():
+            self.set_attribute(NIRFSG_ATTR_REF_CLOCK_SOURCE,
+                               CLK_SRC_MAP[source])
+        else:
+            warnings.warn(f"Unknown clock source {source}."
+                          f" Valid sources are {CLK_SRC_MAP.keys()}. Falling"
+                          f" back to onboard clock.", RuntimeWarning)
+            self.set_attribute(NIRFSG_ATTR_REF_CLOCK_SOURCE,
+                               CLK_SRC_MAP['onboard'])
```
astafan8, line 215: is it desired to have a fallback for this instrument? in case it's not, then the `else` part can be removed, and `CLK_SRC_MAP` can be used in the parameter's validator (via `val_mapping`), and extra information can be added via `docstring` kwarg of the parameter.

qcodes_contrib_drivers/drivers/NationalInstruments/RFSG.py
```diff
@@ -0,0 +1,245 @@
...
+        """
+        self.wrapper.Initiate(self._handle)
+
+    def abort(self):
+        """
+        Stop signal generation and return to the Configuration state.
+        """
+        self.wrapper.Abort(self._handle)
+
+    def ConfigureRF(self, frequency: float, power_level: float,
```
astafan8, line 162: shouldn't this have a python-style name as per recommendation in the README inside NationalInstruments dir?

qcodes_contrib_drivers/drivers/NationalInstruments/RFSG.py
```diff
@@ -0,0 +1,245 @@
...
+            power_level: power level in dBm
+            initiate: if True, call self.initiate after configuring, which
+                starts RF output
+        """
+        self.wrapper.ConfigureRF(self._handle, ViReal64(frequency),
+                                 ViReal64(power_level))
+        if initiate:
+            self.initiate()
+
+    def set_frequency(self, frequency: float, initiate: bool = False):
```
astafan8, line 184: is it desired to have access to frequency (and other settings) BOTH via a parameter and a method? if not, then the recommendation is to leave parameter as the only "interface", and make the method private (prepend with `_`)

qcodes_contrib_drivers/drivers/NationalInstruments/RFSG.py
```diff
@@ -0,0 +1,245 @@
...
+
+        self.add_parameter(name="power_level",
+                           unit="dBm",
+                           label="power level",
+                           get_cmd=partial(self.get_attribute,
+                                           NIRFSG_ATTR_POWER_LEVEL),
+                           set_cmd=self.set_power_level,
+                           )
+
+        self.add_parameter(name="output_on",
```
astafan8, line 117: shall it be renamed to `output_enabled`? that's what the label says, and that's what qcodes also recommends (in its writing drivers page of the docs)

qcodes_contrib_drivers/drivers/NationalInstruments/PXIe_5654.py
```diff
@@ -0,0 +1,40 @@
...
+        self.frequency.vals = Numbers(250e3, 20e9)
+        self.power_level.vals = Numbers(-7, 15)
+
+        # check for amplitude extender and update power limits accordingly
+        model = self.IDN()["model"]
+        if "PXIe-5696" in model:
+            self.power_level.vals = Numbers(-110, 24)
+
+
+NI_PXIe_5654 = NationalInstruments_PXIe_5654
```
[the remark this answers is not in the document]
mgunyho: The long name version is trying to be consistent with the other drivers. Looking around more carefully I now see that the classes in most other drivers where the manufacturer can be abbreviated don't have such names.

I think the naming conventions could be more explicit in the docs, although I understand that the names of the devices and manufacturers themselves are sometimes quite a mess and it's hard to come up with rules that fit every case.

qcodes_contrib_drivers/drivers/NationalInstruments/RFSG.py
mgunyho: Fixed in 2abeac7.

qcodes_contrib_drivers/drivers/NationalInstruments/RFSG.py
mgunyho: True. There exists a function in the driver that takes these parameters, but we ended up not using it.

qcodes_contrib_drivers/drivers/NationalInstruments/RFSG.py
mgunyho: Good point. This will require a bit of search-and replace in our scripts but oh well. Fixed in a504142.

PR #49, opened 2020-04-08
Ni Switch RF switches

mgunyho:
This PR adds a basic driver for RF switches based on the [`niswitch`](https://nimi-python.readthedocs.io/en/latest/niswitch.html) module, and a driver for the PXIe-2597 multiplexer.


qcodes_contrib_drivers/drivers/NationalInstruments/PXIe_2597.py
```diff
@@ -0,0 +1,58 @@
...
+                 reset_device: bool = False, **kwargs):
+
+        if name_mapping is not None:
+            # don't mutate external dict
+            name_mapping = name_mapping.copy()
+            name_mapping["com"] = "com"
+
+        super().__init__(name, resource, name_mapping, reset_device, **kwargs)
+
+        self.channels.lock()
```
astafan8, line 31: why locking channels only here?

mgunyho: I think the idea was that subclasses could customize the channel list if necessary. But you're right, at least currently there's no reason why it's not in the superclass.

PR #50, opened 2020-04-09
Vaunix LDA step attenuator

mgunyho:
This PR contains a driver for Vaunix LDA digital attenuators.

### DIFF
--- a/qcodes_contrib_drivers/drivers/Vaunix/LDA.py
+++ b/qcodes_contrib_drivers/drivers/Vaunix/LDA.py
@@ -7,7 +7,7 @@ created for each one. If the instrument has only one physical channel, no
 channels are created and the parameters will be assigned to this instrument
 instead. The sweep profiles available in the API are not implemented.
 
-Tested with with 64-bit system and
+Tested with 64-bit system and
 - LDA-133
 - LDA-802Q
 
--- a/qcodes_contrib_drivers/drivers/Vaunix/LDA.py
+++ b/qcodes_contrib_drivers/drivers/Vaunix/LDA.py
@@ -1,28 +1,15 @@
 import logging
-from typing import Optional, List, Any, Dict
+from typing import Optional, List, Any, Dict, Callable
 from functools import partial
 from platform import architecture
 import os
+import sys
 import ctypes
 
 from qcodes import Instrument, InstrumentChannel
 from qcodes.utils.validators import Numbers
 
 logger = logging.getLogger(__name__)
-
-# path to dll file
-dll_path = os.path.dirname(os.path.abspath(__file__))
-       
-bitness = architecture()[0]
-if "64bit" in bitness:
-    dll_path = os.path.join(dll_path, "VNX_atten64")
-elif "32bit" in bitness:
-    dll_path = os.path.join(dll_path, "VNX_atten")
-else:
-    raise SystemError("Unkown bitness of system:", bitness)
-DLL = ctypes.cdll.LoadLibrary(dll_path)
-
-DLL.fnLDA_SetTestMode(False)  # Test API without communication
 
 ATT_UNIT = 0.05  # integers returned by the API correspond to 0.05 dB
 FREQ_UNIT = 100_000  # integers returned by the API correspond to 100kHz
@@ -31,10 +18,10 @@ FREQ_UNIT = 100_000  # integers returned by the API correspond to 100kHz
 class LDA(Instrument):
     r"""
     This is the QCoDeS driver for Vaunix LDA digital attenuators.
-    Assumes a 64-it Windows system and that `VNX_atten64.dll` and/or `VNX_atten.dll`
-    is located in the same folder.
-    If the instrument has more than one physical channel, `InstrumentChannel` objects
-    are created for each one. If there is only one physical channel, no channels are created
+    Requires that the DLL that comes with the instrument, `` VNX_atten64.dll``  and/or `` VNX_atten.dll``,
+    is found at ``dll_path``, for 64-bit Windows and 32-bit Windows, respectively.
+    If the instrument has more than one physical channel, `` InstrumentChannel``  objects
+    are created for each one. If the instrument has only one physical channel, no channels are created
     and the parameters will be assigned to this instrument instead.
     The sweep profiles available in the API are not implemented.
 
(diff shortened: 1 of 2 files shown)


qcodes_contrib_drivers/drivers/Vaunix/LDA.py
```diff
@@ -0,0 +1,291 @@
...
+
+        Args:
+            name: Qcodes name for this instrument
+            serial_number: Serial number of the instrument, used to identify
+                it.
+            dll_path: Look for the LDA DLLs in this directory.
+                By default, the directory this file is in.
+            channel_names: Optionally assign these names to the channels.
+            test_mode: If True, communicates with the API but not with the
+                device. For testing purposes.
```
astafan8, line 48: and in the case of test mode, the serial number determines what model of the instrument the API is simulating? if so, it'd be great to mention this here.

qcodes_contrib_drivers/drivers/Vaunix/LDA.py
```diff
@@ -0,0 +1,291 @@
...
+                 **kwargs):
+        r"""
+        QCoDeS Instrument for Vaunix LDA digital attenuators.
+
+        Args:
+            name: Qcodes name for this instrument
+            serial_number: Serial number of the instrument, used to identify
+                it.
+            dll_path: Look for the LDA DLLs in this directory.
+                By default, the directory this file is in.
```
astafan8, line 45: i don't think this is a good default. qcodes-contrib-drivers is published on pypi and users are encouraged to install it from there (as opposed to cloning the repo) - in that case the path to this file will be buries somewhere inside virtualenv/conda environement which is not convenient, and users should not go and add files there. Hence, it's better if the default DLL path is something that the manufacturer imposes, for example, quite a few instruments that i've seen install their DLLs somewhere in program files. If the manufacturer doesn't suggest any path then simple `<user dir>` or `C:/` is much better for a default.

qcodes_contrib_drivers/drivers/Vaunix/LDA.py
```diff
@@ -0,0 +1,291 @@
...
+        r"""
+        Load correct DLL from ``dll_path`` based on bitness of the operating
+        system.
+
+        Args:
+            dll_path: path to the directory that contains the Vaunix LDA DLL.
+                By default, same as the directory of this file.
+        """
+        if dll_path is None:
+            dll_path = os.path.dirname(os.path.abspath(__file__))
```
astafan8, line 105: a good practice is to expose the dll_path as a class attribute, so that users can set it before initializing the instrument AND don't need to pass it to the instrument's init each time.

qcodes_contrib_drivers/drivers/Vaunix/LDA.py
```diff
@@ -0,0 +1,291 @@
...
+            name: parameter name
+            instrument: parent instrument, either LDA or LDA channel
+            dll_get_function: DLL function that gets the value
+            dll_get_function: DLL function that sets the value
+        """
+        super().__init__(name, instrument, **kwargs)
+        self.dll = instrument.root_instrument.dll
+        self.reference = instrument.root_instrument.reference
+        self.dll_get_function = partial(dll_get_function, self.reference)
+        self.dll_set_function = partial(dll_set_function, self.reference)
```
astafan8, line 185: i'd make all of these attributes private in order not to confuse users :)

qcodes_contrib_drivers/drivers/Vaunix/LDA.py
```diff
@@ -0,0 +1,291 @@
...
+    scaling = 0.05  # integers returned by the API correspond to 0.05 dB
+
+    def __init__(self, name: str,
+                 instrument: Union[Vaunix_LDA, LdaChannel],
+                 **kwargs):
+        dll = instrument.root_instrument.dll
+        super().__init__(name, instrument,
+                         dll_get_function=dll.fnLDA_GetAttenuationHR,
+                         dll_set_function=dll.fnLDA_SetAttenuationHR,
+                         vals=Numbers(0, 63),  # replace below
```
astafan8, line 232: can't the validator be prepared before calling `super.__init__`?

qcodes_contrib_drivers/drivers/Vaunix/LDA.py
```diff
@@ -0,0 +1,291 @@
...
+    def get_validator(cls, root_instrument: Vaunix_LDA) -> Optional[Numbers]:
+        """
+        Returns validator for working frequency, if ``root_instrument``
+        supports it. Else returns None.
+        """
+        max_freq = root_instrument.dll.fnLDA_GetMaxWorkingFrequency(
+                    root_instrument.reference) * cls.scaling
+        min_freq = root_instrument.dll.fnLDA_GetMinWorkingFrequency(
+                    root_instrument.reference) * cls.scaling
+        # if feature is not supported, these values will be equal
```
astafan8, line 281: shall then this be the condition instead of "whether `Numbers(min_freq, max_freq)` raises `TypeError` or not"?

qcodes_contrib_drivers/drivers/Vaunix/LDA.py
```diff
@@ -0,0 +1,291 @@
...
+        self.reference = devices.get(self.serial_number, "not found")
+        if self.reference == "not found":
+            raise ValueError(f"LDA with serial number {self.serial_number}"
+                             f" was not found in the system. Found: {devices}")
+
+        self.dll.fnLDA_InitDevice(self.reference)
+
+        num_channels = self.dll.fnLDA_GetNumChannels(self.reference)
+        if num_channels == 1:
+            # don't add Channel objects, add parameters directly instead
```
astafan8, line 77: in order not to repeat this code in LdaChannel, it could be extracted into a function that takes `instrument_for_parameters` and `root_instrument` as arguments.

qcodes_contrib_drivers/drivers/Vaunix/LDA.py
mgunyho: Fixed in 0191808.

qcodes_contrib_drivers/drivers/Vaunix/LDA.py
mgunyho: Fixed in 0191808.

qcodes_contrib_drivers/drivers/Vaunix/LDA.py
mgunyho: Fixed in 0191808.

qcodes_contrib_drivers/drivers/Vaunix/LDA.py
mgunyho: Fixed in a72e36d.

qcodes_contrib_drivers/drivers/Vaunix/LDA.py
mgunyho: Fixed in 5d9a29e.

qcodes_contrib_drivers/drivers/Vaunix/LDA.py
mgunyho: Fixed in 5226eac2.

qcodes_contrib_drivers/drivers/Vaunix/LDA.py
mgunyho: Fixed in 5d9a29e.

qcodes_contrib_drivers/drivers/Vaunix/LDA.py
```diff
@@ -0,0 +1,299 @@
...
+    def _get_dll(self, dll_path: str = None) -> ctypes.CDLL:
+        r"""
+        Load correct DLL from ``dll_path`` based on bitness of the operating
+        system.
+
+        Args:
+            dll_path: path to the directory that contains the Vaunix LDA DLL.
+                By default, use class attribute ``Vaunix_LDA.dll_path``.
+        """
+        Vaunix_LDA.dll_path = dll_path or Vaunix_LDA.dll_path
```
astafan8, line 100: in this case, it might not be smart to set the class attribute from an instance. my idea was that the user either specifies `Vaunix_LDA.dll_path` before initializing OR passes `dll_path` to `__init__`. But it's up to you, what you think is most appropriate in this case.

mgunyho: Fair point, updated in 46a9c13.

PR #52, opened 2020-04-09
Keithley 6430 SMU

mgunyho:
This is a port of an old driver, with a lot of parameters. Not all of them have been tested, but the basic features work. Unfortunately, there's no example notebook.


qcodes_contrib_drivers/drivers/Tektronix/Keithley_6430.py
```diff
@@ -0,0 +1,431 @@
...
+                           vals=Enum(200, 20, 2, 0.2),
+                           )
+        self.add_parameter('sense_resistance_range',
+                           units='Ohm',
+                           get_parser=float,
+                           set_cmd=':SENS:RES:RANG {}',
+                           get_cmd=':SENS:RES:RANG?',
+                           vals=Numbers(2, 2e13),
+                           )
+        self.add_parameter('sense_resistance_ocomp',
```
astafan8, line 192: ?
```suggestion
        self.add_parameter('sense_resistance_offset_comp_enabled',
```

qcodes_contrib_drivers/drivers/Tektronix/Keithley_6430.py
```diff
@@ -0,0 +1,431 @@
...
+        self.sense_mode('VOLT:DC,CURR:DC')
+        # self.set_trigger_cont(True)
+        self.source_current_range(1e-6)
+        self.source_current(0e-6)
+        self.source_current_compliance(1e-6)
+        self.source_voltage_range(200e-3)
+        self.source_voltage_compliance(1e-3)
+        self.digits(7)
+        self.nplc(10)
+        # self.set_averaging(False)
```
astafan8, line 319: perhaps remove commented out lines?

qcodes_contrib_drivers/drivers/Tektronix/Keithley_6430.py
```diff
@@ -0,0 +1,431 @@
...
+        Note that the values may not be valid if sense mode doesn't include
+        them.
+        '''
+        s = self.ask(':FETC?')
+
+        # We don't know what [-2:] are...
+        v, c, r = [float(n) for n in s.split(',')][-5:-2]
+        return (v, c, r)
+
+    def set_trigger_cont(self):
```
astafan8, line 372: perhaps?
```suggestion
    def set_trigger_imm(self):
```

qcodes_contrib_drivers/drivers/Tektronix/Keithley_6430.py
```diff
@@ -0,0 +1,431 @@
...
+from qcodes.utils.validators import Ints, Numbers, Bool, Strings, Enum
+from qcodes.utils.helpers import create_on_off_val_mapping
+import logging
+import warnings
+from functools import partial
+from typing import Tuple
+
+on_off_vals = create_on_off_val_mapping(on_val=1, off_val=0)
+
+
```
astafan8, line 23: has there not been problems with taking snapshot of this instruemnt? if there are, then it might be worth looking to various `snapshot_*` flags that `Parameter` exposes and use them for some of the parameters.

mgunyho:
Thanks for the feedback! I forgot to merge some changes we already had made previously (like chaning `output_on` to `output_enabled`). There also was in fact an example notebook. All the changes proposed above should be implemented as of 999c850.

qcodes_contrib_drivers/drivers/Tektronix/Keithley_6430.py
astafan8: @mgunyho what about this question?

mgunyho: I'm not sure I entirely understand the question. Is it related to the comment

```
# NOTE that using 1/0 integer values will also work implicitly due to `hash(True) == hash(1)`/`hash(False) == hash(0)`
```
in `create_on_off_val_mapping`? We use `snapshot_value=False` in some of the parameters to supprress some warnings.

astafan8: > Is it related to the comment

nope :)

> We use snapshot_value=False in some of the parameters to supprress some warnings.

what warnings are being suppressed?

it is important. If these warnings can be addressed by careful use of snapshot related flags (`snapshot_value` , `snapshot_get`, `snapshot_exclude`) and some customization of parameter's set/get methods, then - great - because this will result in a more completed an meaningful snapshot.

-----------

the reason i initially asked the question is that this instrument seemingly has a few modes where in one mode values of some parameters can not be "gotten", and in another mode this is true for some other parameters. Indeed, the simple solution is just to exclude all of those parameters from the snapshot - but then the benefit of the snapshot is lost. So, i'm just reminding of a concern here, but it's up to you how to address it in this driver.

Jajiko: You were right, the snapshot works just fine with `snapshot_get=False` and retrives the cached value if there is one. We probably had excluded those to suppress warnings in earlier development stage...
Anyway, this was fixed in d0dbdea

PR #58, opened 2020-05-29
Small fixes to Vaunix LDA driver

mgunyho:
Improved error handling in LDA initialization and close, and add set_parser for the attenuation parameter.

### DIFF
--- a/qcodes_contrib_drivers/drivers/Vaunix/LDA.py
+++ b/qcodes_contrib_drivers/drivers/Vaunix/LDA.py
@@ -112,7 +112,7 @@ class Vaunix_LDA(Instrument):
         elif "32bit" in bitness:
             full_path = os.path.join(path, "VNX_atten")
         else:
-            raise OSError("Unknown bitness of system:", bitness)
+            raise OSError(f"Unknown bitness of system: {bitness}")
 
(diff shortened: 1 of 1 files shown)


qcodes_contrib_drivers/drivers/Vaunix/LDA.py
```diff
@@ -51,8 +52,8 @@ def __init__(self, name: str,
                 (serial:55102). Does not communicate with physical devices. For
                 testing purposes.
         """
+        begin_time = time.time()
```
GateBuilder, line 55: Why do we need this?

mgunyho: Otherwise the `connect_message` displays 0.00 seconds, it's mostly just cosmetic

GateBuilder: Thanks, fair enough. I believe you already tested this improved version with a real instrument?

mgunyho: Yes, that's how we found these bugs

GateBuilder: Thanks.

Issue #61, opened 2020-07-18
NationalInstruments drivers missing when installing using pip


mluessi:
For some reason, the NationalInstruments folder and the drivers it contains are missing for me.

I tried the following:

Install version 0.5.0:

    pip install qcodes_contrib_drivers

Install latest version from github:

    pip install git+https://github.com/QCoDeS/Qcodes_contrib_drivers.git

In both cases, the drivers folder does not contain the NationalInstruments folder. Any ideas?

![image](https://user-images.githubusercontent.com/647005/87854519-6f535200-c8c7-11ea-91c0-e9df5dc99c84.png)

jenshnielsen:
It looks like the NationalInstruments folder is missing an empty ``__init__.py`` file. @mluessi it would be great if you can create a pr to add that file

jenshnielsen:
fixed by #62

PR #62, opened 2020-07-28
Add init file for NationalInstruments

Akshita07:
Fixes issue #61 

Add empty __init__.py file for NationalInstruments folder

@WilliamHPNielsen

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -104,6 +104,8 @@ nbsphinx_execute = 'always'
 autodoc_mock_imports = ['spirack',
                         'pyspcm',
                         'keysightSD1',
+                        'nidaqmx',
+                        'niswitch',
                         'zhinst']
 
 # we allow most types from the typing modules to be used in
@@ -138,6 +140,7 @@ nitpick_ignore = [('py:class', 'Optional'),
                   ('py:class', 'json.encoder.JSONEncoder'),
                   ('py:class', 'SPI_rack'),
                   ('py:class', 'spirack.SPI_rack'),
+                  ('py:class', 'ViAttr'),
                   ('py:class', 'unittest.case.TestCase'),
                   ('py:class', 'builtins.AssertionError'),
                   ('py:class', '_ctypes.Structure'),
--- a/qcodes_contrib_drivers/drivers/NationalInstruments/PXIe_2597.py
+++ b/qcodes_contrib_drivers/drivers/NationalInstruments/PXIe_2597.py
@@ -7,7 +7,7 @@ class NI_PXIe_2597(NI_Switch):
     r"""
     QCoDeS driver for National Instruments RF switch PXIe-2597. The device
     connects the common "com" port to any of the 6 other ports, labeled
-    "ch1"..."ch6" by default. Use the ``name_mapping `` argument to alias the
+    "ch1"..."ch6" by default. Use the ``name_mapping`` argument to alias the
     channel names.
 
     Args:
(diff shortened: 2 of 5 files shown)


Akshita07:
@mgunyho @Jajiko to solve issue #61 this pr was created and an empty init file was added. Following that some minor mypy issues are coming. Is this something you might want to look into?

mgunyho:
I can try to look into it

mgunyho:
I can't reproduce the errors in `RFSG.py`. I kind of understand where they are coming from, but for me mypy doesn't report them as errors (running with `mypy --config-file=../../../mypy.ini RFSG.py` in the NationalInstruments folder). The other ones I've fixed, I'll test if they really work and publish them then.

Akshita07:
@mgunyho please make sure that your locally installed mypy version is same as mentioned in https://github.com/QCoDeS/Qcodes/blob/master/test_requirements.txt file.

mgunyho:
I figured it out, I had to run `mypy qcodes_contrib_drivers` in the root folder. I couldn't test everything today, I'll get back to it next week.

mgunyho:
I finally got around to testing this with the hardware, sorry for the delay. I've pushed the latest version to https://github.com/mgunyho/Qcodes_contrib_drivers/commit/23a7e763e390d9f6c8e7377aacf020443ab956fa

I had to override `__getattr__` and `__setattr__` in `NIDLLWrapper` to stop mypy from complaining, do you think that's a good idea / does it cause too much confusion?

jenshnielsen:
@mgunyho I think its better to add ignore type comments to silence the issues https://mypy.readthedocs.io/en/stable/common_issues.html#spurious-errors-and-locally-silencing-the-checker

mgunyho:
that feels a bit hackish to me, as you would have to add an ignore comment every time you register or access a wrapped method

jenshnielsen:
@mgunyho This is of cause to a certain extend a matter of personal preference but I would not call it hacking but rather honerst. 
 There is a very good reason why MyPy warns when you dynamically create attributes outside `__init__`. As far as possible that should always be avoided. You have found a way that tricks mypy into not giving that warning but there is no guarantee that this will be the case for any future version of mypy which may impose stricter checks. Furthermore its hard to understand why this works as the standard library defines the same signature for `__setattr__` https://github.com/python/typeshed/blob/master/stdlib/2and3/builtins.pyi#L100 

It's therefor hackish in my opinion to rely on overwriting `__setattr__` and `__getattr__` like this. Having to add an ignore comment for every variable is annying but makes it clear that you are aware of the issues in the design of the code.

mgunyho:
Hmm, that is fair enough, and indeed it is confusing why this changes the behavior compared to the standard library stubs.

Do you think setting attributes outside init is fine here (`self.wrapper.Function = self.wrapper.wrap_dll_function(...)`)? I originally had it so that `wrap_dll_function` would do `setattr(self, name, func)`, but I changed to the current version because it's much more explicit and easier to understand. The previous version had similar issues with mypy complaining about non-existent attributes, both when registering and accessing them.

Nevertheless, I've removed the `__get/setattr__` overrides and added `type: ignore` as of commit https://github.com/mgunyho/Qcodes_contrib_drivers/commit/1320ab2bdae2e2533e1dbaba2b0c081c9a8b802e

jenshnielsen:
I would normally prefer not to define attribues outside init but I think its ok in this case since its relatively clear that this is correct and done on purpose and not due to a typo so lets go ahead with that

jenshnielsen:
@mgunyho are you interested in putting those changes into a pull request ?

mgunyho:
My branch (https://github.com/mgunyho/Qcodes_contrib_drivers/commit/7a616d0d11df2643be539546124a3e1466fb98a8) is on top of this one, so @Akshita07 could just pull my changes to this branch, but I can also make a separate PR if you wish

jenshnielsen (approved):

PR #67, opened 2020-07-31
Adding an empty __init__.py to Vaunix folder driver

FarBo:


jenshnielsen:
@FarBo This check needs to be at the top level unindented see here https://mypy.readthedocs.io/en/stable/common_issues.html?highlight=platform#python-version-and-system-platform-checks

jenshnielsen:
thanks @FarBo

jenshnielsen (approved):

PR #73, opened 2020-09-04
Feature/Rohde&Schwarz HMP series

MarceloSemO:
Driver for the R&S HMP series (similar to driver for HMC series, but with some subtle differences)

### DIFF
--- a/qcodes_contrib_drivers/drivers/RohdeSchwarz/private/HMP.py
+++ b/qcodes_contrib_drivers/drivers/RohdeSchwarz/private/HMP.py
@@ -1,0 +1,114 @@
+from qcodes import VisaInstrument, validators as vals
+from qcodes import InstrumentChannel, ChannelList
+from functools import partial
+
+
+class RohdeSchwarzHMPChannel(InstrumentChannel):
+    def __init__(self, parent, name, channel):
+        super().__init__(parent, name)
+        self.channel = channel
+        self.max_current = self.get_max_current()
+
+
+        self.add_parameter("set_voltage",
+                           label='Target voltage output',
+                           set_cmd=partial(self.send_cmd, "set_voltage"),
+                           get_cmd=partial(self.send_cmd, "set_voltage", None),
+                           get_parser=float,
+                           unit='V',
+                           vals=vals.Numbers(0, 32.050)
+                           )
+        self.add_parameter("set_current",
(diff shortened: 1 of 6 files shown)


qcodes_contrib_drivers/drivers/RohdeSchwarz/private/HMP.py
```diff
@@ -0,0 +1,118 @@
...
+        if self.parent.model_no > 4000:
+            return 10
+        elif self.parent.model_no == 2020 and self.channel == 1:
+            return 10
+        return 5
+
+    def send_cmd(self, param, value):
+        self.write(f"INSTrument:NSELect {self.channel:d}")
+        if value is None:
+            return self.ask(f"{self.get_scpi_command(param)}?")
```
jenshnielsen, line 65: I would normally not make get_scpi_command a method. I would just make the dictionary a private member variable on the class and do something like `self.ask(f"{self._commands[param]}?")`
Alternatively you could pass the command in via another argument to send_cmd. That would have the advantage of including the command with the parameter definition but at the cost of slight repetition.

qcodes_contrib_drivers/drivers/RohdeSchwarz/private/HMP.py
```diff
@@ -0,0 +1,118 @@
...
+        commands = {
+            "set_voltage": "SOURce:VOLTage:LEVel:IMMediate:AMPLitude",
+            "set_current": "SOURce:CURRent:LEVel:IMMediate:AMPLitude",
+            "state": "OUTPut:STATe",
+            "voltage": "MEASure:SCALar:VOLTage:DC",
+            "current": "MEASure:SCALar:CURRent:DC",
+        }
+        return commands.get(param)
+
+    def get_power(self):
```
jenshnielsen, line 81: I don't think the user should ever need to call this directly so I would make it protected with a name like `_get_power`

jenshnielsen (approved):
Thanks, I left a few minor suggestions for potential improvements

MarceloSemO:
@jenshnielsen, your suggestions are implemented now and I have rebased the branch to incorporate the recent merges into the master branch. I think this feature is now also ready for merging into the master branch :)

jenshnielsen:
Thanks @MarceloSemO I will merge it now

PR #74, opened 2020-09-14
Attocube ANC350 driver

lankes-fzj:
Driver for *Attocube ANC350* piezo controller.

It supports version 3 (`anc350v3.dll`) and 4 (`anc350v4.dll`) of the driver DLL. Version 2 (`hvpositionerv2.dll`, not compatible with version 3) is not supported as it couldn't be tested.


FarBo:
Hello @lankes-fzj 
Thank you a lot for the PR. I will try to start the review as soon as possible.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,520 @@
+﻿from qcodes import Instrument, Parameter
+from qcodes.instrument.channel import InstrumentChannel, ChannelList
+from qcodes.utils.validators import Numbers
+from qcodes_contrib_drivers.drivers.Attocube.ANC350Lib import ANC350LibActuatorType, ANC350v3Lib, ANC350v4Lib
+from typing import Any, Callable, Dict, Optional, Union
+
+
+class ANC350OutputParameter(Parameter):
```
FarBo, line 8: Qcodes already supports callable get_cmd and set_cmd, which are wrapped in get_raw and set_raw, respectively. So, I am not sure why there is a need for this class.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,520 @@
...
+
+    def __init__(self, parent: "ANC350", name: str, axis: int):
+        super().__init__(parent, name)
+
+        self._axis = axis
+
+        self.add_parameter("position",
+                           label="Position",
+                           get_cmd=self._get_position,
+                           set_cmd=False,
```
FarBo, line 63: The default of set_cmd is False, so this line can be removed.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,520 @@
...
+                           unit="mm or m°")
+
+        self.add_parameter("frequency",
+                           label="Frequency",
+                           get_cmd=self._get_frequency,
+                           set_cmd=self._set_frequency,
+                           unit="Hz")
```
FarBo, line 70: Validator for this parameter is needed

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,520 @@
...
+    def __init__(self, parent: "ANC350", name: str, axis: int):
+        super().__init__(parent, name)
+
+        self._axis = axis
+
+        self.add_parameter("position",
+                           label="Position",
+                           get_cmd=self._get_position,
+                           set_cmd=False,
+                           unit="mm or m°")
```
FarBo, line 64: Validator for this parameter is needed

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
lankes-fzj: Yes, I know, but I didn't manage to make the optional keyword argument (`auto_off`) working. The problem was, that the `set_cmd` is wrapped by `qcodes.utils.command.Command` which doesn't forward the keyword-arguments.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
lankes-fzj: Really? I thought that only settable parameters need a validator.

FarBo: My bad. No need here.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
FarBo: This class is used further down in "output" in line 133 and nowhere else (correct me if I'm wrong). For setting that parameter, you have used set_cmd=self._set_output, which is defined in line 429. This function itself takes auto_off argument. So, I am wondering what happens if you remove param_type=ANC350OutputParameter from "output". What kind of error message do you get?

jenshnielsen: @FarBo I am not sure if that is possible when set takes an extra argument. Look at the code here which wraps `set_cmd` https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument/parameter.py#L1126

However, I would suggest that it perhaps makes more sense for auto_off to be a parameter that is set independetly of the set command?

lankes-fzj: I've found another way to realize the output-parameter. It now is a normal Parameter, with three instead of two possible values: "on", "off" and "auto"

FarBo: Sounds good. Could you please share the manual of the instrument as well?

lankes-fzj:
> Sounds good. Could you please share the manual of the instrument as well?

Here you go! [Manual_ANC350_v4.0.11.pdf](https://github.com/QCoDeS/Qcodes_contrib_drivers/files/5408735/Manual_ANC350_v4.0.11.pdf)

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
lankes-fzj: Yes, I've uploaded it to the main conversation: https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/74#issuecomment-712824839

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,510 @@
...
+
+    def __init__(self, parent: "ANC350", name: str, axis: int):
+        super().__init__(parent, name)
+
+        self._axis = axis
+
+        self.add_parameter("position",
+                           label="Position",
+                           get_cmd=self._get_position,
+                           unit="mm or m°")
```
FarBo, line 47: The position unit is um or m°. Is there any reason to use mm?

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,510 @@
+﻿from qcodes import Instrument, Parameter
+from qcodes.instrument.channel import InstrumentChannel, ChannelList
+from qcodes.utils.validators import Numbers
```
FarBo, line 3: Ints should be imported here

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,510 @@
...
+        self.add_parameter("position",
+                           label="Position",
+                           get_cmd=self._get_position,
+                           unit="mm or m°")
+
+        self.add_parameter("frequency",
+                           label="Frequency",
+                           get_cmd=self._get_frequency,
+                           set_cmd=self._set_frequency,
+                           vals=Numbers(),
```
FarBo, line 53: vals=Ints(1, 5000)
In the manual, it is stated that the Frequency can be between 1 Hz to 5 kHz for one axis and between 1 Hz to 2 kHz for all axis, with the resolution of 1 Hz.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
lankes-fzj: The DLL returns ° or m, depending on the current actuator type (see `anc350res.h` in [inc.zip](https://github.com/QCoDeS/Qcodes_contrib_drivers/files/5408772/inc.zip)). Our customer preferred to use mm and m° instead, because they also use these units internally and thus do not confuse the different units.

If you prefer other units, I can change it.

FarBo: I see. It is a bit confusing, though. dll output should conform the units stated in the manual. To me, it is strange to have the position in the order of meter. Micrometer seems to be a correct unit.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,510 @@
...
+                           unit="Hz")
+
+        self.add_parameter("amplitude",
+                           label="Amplitude",
+                           get_cmd=self._get_amplitude,
+                           set_cmd=self._set_amplitude,
+                           vals=Numbers(0, 70),
+                           unit="V")
+
+        self.add_parameter("status",
```
FarBo, line 63: Just a suggestion: to change the name to axis_status and the label to Axis Status

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
lankes-fzj: Ok, then I would suggest to use um and u°, because then I can use the same conversion factor for both units. So I don't have to check the actuator every time I want to read the position.

Otherwise I had to use a different factor for each unit when using um and m° and I am afraid that checking the actuator will go wrong and the position will be wrong by a factor of 1000.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
lankes-fzj: I don't think, that this is necessary. When using the instrument later, you will call `instrument.x_axis.status`. So, you have to type less and it's still clear that it's the axis-status.

lankes-fzj:
I'm not sure, what's wrong now. The tests are failing, but it doesn't look like it was my fault.

It says the test of Rohde Schwarz HMC804x failed, because it couldn't import `pyvisa-sim`. Do you have any idea why this is happening? Has the CI-configuration changed?

Here is an extract from the log of "Pytest on Linux":
```
tests/test_rohdeschwarz_HMC804x.py:14: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
drivers/RohdeSchwarz/private/HMC804x.py:72: in __init__
    super().__init__(name, address, **kwargs)
/usr/share/miniconda/envs/qcodes/lib/python3.7/site-packages/qcodes/instrument/visa.py:87: in __init__
    raise e
/usr/share/miniconda/envs/qcodes/lib/python3.7/site-packages/qcodes/instrument/visa.py:83: in __init__
    self.set_address(address)
/usr/share/miniconda/envs/qcodes/lib/python3.7/site-packages/qcodes/instrument/visa.py:113: in set_address
    resource_manager = visa.ResourceManager(self.visalib)
/usr/share/miniconda/envs/qcodes/lib/python3.7/site-packages/pyvisa/highlevel.py:1553: in __new__
    visa_library = open_visa_library(visa_library)
/usr/share/miniconda/envs/qcodes/lib/python3.7/site-packages/pyvisa/highlevel.py:1522: in open_visa_library
    cls = get_wrapper_class(wrapper)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

backend_name = 'sim'

    def get_wrapper_class(backend_name):
        """Return the WRAPPER_CLASS for a given backend.
    
        :rtype: pyvisa.highlevel.VisaLibraryBase
        """
        try:
            return _WRAPPERS[backend_name]
        except KeyError:
            if backend_name == 'ni':
                from .ctwrapper import NIVisaLibrary
                _WRAPPERS['ni'] = NIVisaLibrary
                return NIVisaLibrary
    
        try:
            pkg = __import__('pyvisa-' + backend_name)
            _WRAPPERS[backend_name] = cls = pkg.WRAPPER_CLASS
            return cls
        except ImportError:
>           raise ValueError('Wrapper not found: No package named pyvisa-%s' % backend_name)
E           ValueError: Wrapper not found: No package named pyvisa-sim

/usr/share/miniconda/envs/qcodes/lib/python3.7/site-packages/pyvisa/highlevel.py:1470: ValueError
- generated xml file: /home/vsts/work/1/s/qcodes_contrib_drivers/test-results.xml -

----------- coverage: platform linux, python 3.7.7-final-0 -----------
Coverage HTML written to dir htmlcov
Coverage XML written to file coverage.xml

=========================== short test summary info ============================
ERROR tests/test_rohdeschwarz_HMC804x.py::test_init - ValueError: Wrapper not...
=================== 12 passed, 1 skipped, 1 error in 19.26s ====================
```

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
FarBo: From the manual, which shows their manufacturer software menus, all positions are stated as um and m° with the resolution of nm and u°. It means that we can't rely on the manual for all units and units depends on the output of dll file, right? The internal resolution is 1 nm or 1 µ°, so how you could use the same conversion for linear and rotational positions?

FarBo:
Could you run pytest on your local cloned repository? Let me know if it passes or fails.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
lankes-fzj: The DLL returns positions in units of m or °. It's some time ago, but I'm pretty sure that we've tested it. E.g. the function `_get_position` currently doesn't know, which unit was returned. It just sees the float-value.. So the best would be to use a constant factor for conversion, e.g. 10^3 to convert units into mm and m° or 10^6 to convert units into µm and µ°.

If you want to use um and m° instead, I had to use a factor of 10^6 for converting m in um and 10^3 to convert ° in m°. Of course, this is possible and not very complicated, but I am not sure if it is worth it.

The advantage would be that you have units that are the same as in LabView, but the problem would be, that we then need two function-calls to get the position (+ actuator type), which might make the driver a bit slower. And I'm afraid, what happens, when the actuator type changes during measurement. If it first was linear, the position unit is m. When reading the position, the result is multiplied by 10^6 to convert it into µm. If the actuator type changes and the unit is now °, it has to be mutliplied by 10^3 to convert it into m°. But what if the actuator changes during measurement and the driver doesn't notice. Then it might happen that it uses the wrong conversion factor and returns mm or µ°, what is not expected by the user.

Additionally, I think it is a bit overkill to do so much background conversions, just to have other units. Since the parameters are given as floats anyway, the internal precision of nm and µ° can still be achieved.

But if you prefer using µm and m°, I can implement it like this. I just wanted you to know my concerns.

lankes-fzj:
> Could you run pytest on your local cloned repository? Let me know if it passes or fails.

It succeeds, but if I uninstall `pyvisa-sim`, I get the same error. I think, `pyvisa-sim` wasn't installed automatically, and I had to install it myself some time ago. Maybe it's sufficient to add `pyvisa-sim` to the requirements in `setup.py`?

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
FarBo: OK. I am not saying we have to have specific units, although it is recommended to conform to the manual. The only point is that the instrument sets/ gets correct location and resolution. So if you have tested the instrument already and you know what the returned numbers and your correct conversion are, then you can program it like that and inform users which units are used as you did in the code.

The important point is to define correct resolution for 2 parameters that can be set: 'Target Range' and 'Target Position'. You should be careful as you used 1e-6 for both units.

FarBo:
pyvisa-sim is part of the test_requirements.txt in qcodes, which can be installed to the environment by pip install -r test_requirements.txt. You may installed it directly on your local machine. We have changed pyvisa and pyvisa-sim versions yesterday and you can find the versions in the requirements.txt and test_requirements.txt, respectively in qcodes repository.

lankes-fzj:
> pyvisa-sim is part of the test_requirements.txt in qcodes, which can be installed to the environment by pip install -r test_requirements.txt. You may installed it directly on your local machine. We have changed pyvisa and pyvisa-sim versions yesterday and you can find the versions in the requirements.txt and test_requirements.txt, respectively in qcodes repository.

I've found the problem in the log of "Install environment, qcodes" in Linux CI pipeline. Here are some lines from the log:

- [line 951](https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=16839&view=logs&j=347123c7-c6ed-5812-0c14-06a064703b24&t=c320dcc8-0625-5b99-5010-296b08bf9782&l=951): `Successfully installed PyVISA-1.11.1 PyVISA-sim-0.4.0 [...]`
  The correct version of pyvisa (1.11.1) is successfully installed
- [line 1097](https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=16839&view=logs&j=347123c7-c6ed-5812-0c14-06a064703b24&t=c320dcc8-0625-5b99-5010-296b08bf9782&l=1097): `Collecting pyvisa<1.11,>=1.9.1`
  Then a requirement needs an earlier pyvisa version than 1.11
- [line 1182](https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=16839&view=logs&j=347123c7-c6ed-5812-0c14-06a064703b24&t=c320dcc8-0625-5b99-5010-296b08bf9782&l=1182): `Successfully uninstalled PyVISA-1.11.1`
  pyvisa 1.11.1 is uninstalled
- **[line 1188](https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=16839&view=logs&j=347123c7-c6ed-5812-0c14-06a064703b24&t=c320dcc8-0625-5b99-5010-296b08bf9782&l=1188): `pyvisa-sim 0.4.0 requires pyvisa>=1.11.0, but you'll have pyvisa 1.10.1 which is incompatible.`**
  **pyvisa-sim cannot be installed as it's not compatible with pyvisa 1.10.1 (requires initially installed version 1.11.1)**
- [line 1189](https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=16839&view=logs&j=347123c7-c6ed-5812-0c14-06a064703b24&t=c320dcc8-0625-5b99-5010-296b08bf9782&l=1189): `Successfully installed pyvisa-1.10.1 qcodes-0.18.0 qcodes-contrib-drivers`
  pyvisa 1.10.1 is installed

I didn't find out, where the requirement from line 1097 (`<1.11,>=1.9.1`) comes from, but I think that's the problem.

jenshnielsen:
Just to clarify.

At the moment qcodes_contrib_drivers is tested with the latest release of qcodes (0.18) That depends on pyvisa < 1.11 and will not work with pyvisa 1.11 further more it needs a fork of pyvisa-sim for the tests to pass correctly.

However the testing dependencies installed (requirements.txt) are grabed from qcodes master which supports pyvisa 1.11 and pyvisa-sim 0.4.0 this means that a non working version combination of pyvisa-sim and pyvisa will be installed. 

The ci job should be updated to not install test reqirements from qcodes master. However the test job should start working again as soon as the next version of qcodes is released within the next few days

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,510 @@
...
+                           get_cmd=None,
+                           set_cmd=self._set_target_range,
+                           vals=vals.Numbers(),
+                           unit="mm or m°")
+
+        self.add_parameter("actuator",
+                           label="Actuator",
+                           get_cmd=None,
+                           set_cmd=self._set_actuator,
+                           vals=vals.Numbers(0, 255))
```
FarBo, line 85: Are you sure this can be both integer and float? To me, it is most likely to be only integer (vals.Ints(0, 255)), but I couldn't find it explicitly in the manual.

lankes-fzj: Yes, you're right. I will change it.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,510 @@
...
+        self.add_parameter("capacitance",
+                           label="Capacitance",
+                           get_cmd=self._get_capacitance,
+                           unit="nF")
+
+        if self._parent._version_no >= 4:
+            voltage_get: Optional[Callable] = self._get_voltage
+        else:
+            voltage_get = None
+
```
FarBo, line 104: A comment to say the acceptable voltage resolution is 0.001 V.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,510 @@
...
+                           vals=vals.Numbers(0, 70),
+                           unit="V")
+
+        self.add_parameter("output",
+                           label="Output",
+                           val_mapping={False: 0,
+                                        True: 1,
+                                        "off": 0,
+                                        "on": 1,
+                                        "auto": 2},
```
FarBo, line 118: Could you point out where you found 'auto' that can be mapped to 2?

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
lankes-fzj: It's already mentioned in the docstring of function `_set_voltage` (see line 310), but I can also add it here.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
lankes-fzj: It's described in a comment above: https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/74#discussion_r508315934

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,510 @@
...
+
+    _direction_mapping: Dict[Any, bool] = {
+        "forward": False,
+        "backward": True,
+        +1: False,
+        -1: True
+    }
+
+    @classmethod
+    def _map_direction_parameter(cls, backward: Optional[Union[bool, str, int]]) -> bool:
```
FarBo, line 179: A None is missing here: Optional[Union[bool, str, int]]=None

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,510 @@
...
+    def stop_continuous_move(self):
+        """Stops continuous motion in forward or backward direction."""
+        self._parent._lib.start_continuous_move(self._parent._device_handle, self._axis, False,
+                                                False)
+
+    _direction_mapping: Dict[Any, bool] = {
+        "forward": False,
+        "backward": True,
+        +1: False,
+        -1: True
```
FarBo, line 175: Shouldn't be 0:False and 1:True?

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,510 @@
...
+        if backward is None:
+            return False
+        if backward in [False, True]:
+            return bool(backward)
+
+        if backward in cls._direction_mapping:
+            return cls._direction_mapping[backward]
+
+        raise ValueError("Unexpected value for argument `backward`. Allowed values are: {}".format(
+            [None, False, True, *cls._direction_mapping.keys()]))
```
FarBo, line 189: This format is preferred:
raise ValueError(f"Unexpected value for argument `backward`. Allowed values are:"
                         f"{[None, False, True, *_direction_mapping.keys()]}")

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,510 @@
...
+        Args:
+            steps: Number of steps to move. The sign indicates the moving direction (+: forward,
+                   -: backward)
+        """
+        backward = (steps < 0)
+        
+        for i in range(abs(steps)):
+            self.single_step(backward)
+
+    def start_continuous_move(self, backward: Optional[Union[bool, str, int]] = None):
```
FarBo, line 152: Please annotate the return value

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,510 @@
...
+        Args:
+            backward: Step direction forward (False) or backward (True). Beside True/False, you can
+                      set the direction to "forward"/"backward" or +1/-1 (default: forward or False)
+        """
+        backward = self._map_direction_parameter(backward)
+
+        self._parent._lib.start_continuous_move(self._parent._device_handle, self._axis, True,
+                                                backward)
+
+    def stop_continuous_move(self):
```
FarBo, line 166: Please annotate the return value

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
lankes-fzj: I know this looks a little confusing. The idea behind it was to make the handling of the driver as intuitive as possible:
- `True`/`False` is to use it as a keyword argument (e.g. `instr.single_step(backward=True)`)
- `"forward"`/`"backward"` for use as (readable) position argument (e.g. `instr.start_continuous_move("backward")`)
- `+1`/`-1` have been added to make the use of these functions similar to the use of `move_multiple_steps` where positive values go forward and negative values go backward.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,510 @@
...
+        Starts continuous motion in forward or backward direction.
+        Other kinds of motion are stopped.
+
+        Args:
+            backward: Step direction forward (False) or backward (True). Beside True/False, you can
+                      set the direction to "forward"/"backward" or +1/-1 (default: forward or False)
+        """
+        backward = self._map_direction_parameter(backward)
+
+        self._parent._lib.start_continuous_move(self._parent._device_handle, self._axis, True,
```
FarBo, line 163: I suppose choosing True and False for start argument is also intentional and under control because the header file says start (1) and stop (0)

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,510 @@
...
+        Sets the amplitude parameter for an axis
+
+        Args:
+            amplitude: Amplitude in Volts [V] (internal resolution is 1mV)
+        """
+        self._parent._lib.set_amplitude(self._parent._device_handle, self._axis, amplitude)
+
+    def _get_status(self) -> Dict[str, bool]:
+        """
+        Reads status information about an axis
```
FarBo, line 287: How does the user know which axis this status belong to? Is it returning the current axis? Then shouldn't we return that axis number?

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,510 @@
...
+        self._parent._lib.set_dc_voltage(self._parent._device_handle, self._axis, voltage)
+
+    def _set_target_position(self, target: float) -> None:
+        """
+        Sets the target position for automatic motion.
+        For linear type actuators the position unit is mm, for goniometers and rotators it is m°.
+
+        Args:
+            target: Target position in millimeters [mm] or millidegrees [m°]. Internal resolution is
+                    1 nm or 1 µ°.
```
FarBo, line 321: The internal resolution is nm or µ° if you convert linear and rotational positions with different conversions. Since you have converted with the same factor to mm and m°, shouldn't the resolution to be µm and µ°?

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,510 @@
...
+                                              target * 1e-3)
+
+    def _set_target_range(self, target_range: float) -> None:
+        """
+        Sets the range around the target position where the target is considered to be reached.
+        For linear type actuators the position unit is mm, for goniometers and rotators it is m°.
+
+        Args:
+             target_range: Target range in millimeters [mm] or millidegrees [m°]. Internal
+                           resolution is 1 nm or 1 µ°.
```
FarBo, line 335: The same argument as above can be also applicable here.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
lankes-fzj: Because of object orientation. This function belongs to class `Anc350Axis`, which represents a single axis. Which axis is represented depends on which axis number (and name) is passed in the constructor. So the user can check e.g. `Anc350Axis.name` or `short_name` if he wants to know which axis he is working with. Additionally, I'm pretty sure that most users only use the top-level instrument `Anc350`. So, the user would call e.g. `anc350.x_axis.status()`. Thus it's pretty clear that the user is requesting the status of `x_axis`.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
lankes-fzj: The internal resoultion is *internal*. It has nothing to do, with the units returned by this function. The internal resolution indicates how fine the instrument itself can set and measure positions *internally*. That means the instrument can set and measure position down to a precision of 1 nm (linear) or 1 µ° (radiator, goniometer). But those are not the units, the instruments interface uses. The DLL returns/expects meters and degrees. To use values that are a bit easier to handle, it was decided to use mm and °m, but this has nothing to do with the internal resolution.

lankes-fzj: The reason why the internal resolution is written in the documentation string is that the user knows the precision of the instrument. Additionally, he also knows that it brings no advantage to specify values in pm or n° precision, because it exceeds the instruments maximum resolution/precision.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
lankes-fzj: I'm not quite sure, what you are telling me with that?

Do you want me to replace the start- and stop-functions with a parameter that is set to True for starting and False for stopping?

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
FarBo: Thanks

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
FarBo: It is fine to have method for this purpose.
The confusion is coming from the header file, when it says start (1) and stop (0). To me, it looks like that 1 is interpreted as start in the DLL, while you used True for the same purpose. So, If you say using True and False is the correct approach, it is completely fine.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
FarBo: I was thinking about the decimal numbers, which raised that question. I am aware of the meaning of resolution and unit on the code side, but thanks for complete explanation. Basically:
we can set for example 2.352367 (in the user side, the unit of this number is mm), which according to the internal precision, it is understandable for the DLL. However, for the rotational position, only first 3 decimals are understandable by the DLL.
My question was about the convenience on the user side, where both linear and rotational position to take the same amount of decimal numbers as shown in the manual.
Sorry if the question was not that clear. But, according to the previous discussions about the conversion factor, there is no need to change anything in this code.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
lankes-fzj: Oh, ok. That is no problem. When you take a look at the `start_continuous_move`-function of `ANC350v3Lib`, you will see that the boolean parameter `start` is converted into an `int32` before it's passed to the DLL-function. So, in the end, the DLL-function is called with 0/1 instead of True/False.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,510 @@
...
+
+        Args:
+            actuator: Actuator selection (0..255)
+        """
+        old_actuator_type = self._get_actuator_type()
+        self._parent._lib.select_actuator(self._parent._device_handle, self._axis, actuator)
+
+        self._update_position_unit(old_actuator_type)
+
+    def _update_position_unit(self, old_actuator_type: Optional[ANC350LibActuatorType] = None):
```
FarBo, line 355: Please annotate this function.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,511 @@
...
+                           label="Actuator Name",
+                           get_cmd=self._get_actuator_name)
+
+        self.add_parameter("capacitance",
+                           label="Capacitance",
+                           get_cmd=self._get_capacitance,
+                           unit="nF")
+
+        if self._parent._version_no >= 4:
+            voltage_get: Optional[Callable] = self._get_voltage
```
FarBo, line 101: I think you forgot to run the function _get_voltage here. Shouldn't be: voltage_get: Optional[Callable] = self._get_voltage()

lankes-fzj: No, the function-object is assigned here, not the function's return value.

FarBo: OK. Then , how is this function object useful in the get_cmd of the voltage parameter, below? I think the user expects to get the voltage value when types voltage() or voltage.get().

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,511 @@
...
+
+class ANC350(Instrument):
+    """
+    Qcodes driver for the ANC350
+
+    Args:
+        name: the name of the instrument itself
+        library: library that fits to the version of the device and provides the appropriate dll
+                 wrappers
+        inst_num: Sequence number of the device to connect to (default: 0, the first device found)
```
FarBo, line 460: I think it should be int_no here

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,511 @@
+﻿from qcodes import Instrument
```
FarBo, line 1: It is recommended to use the full path for qcodes modules:
from qcodes.instrument,base import Instrument

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
lankes-fzj: The function-object is passed to `add_parameter` as argument `get_cmd`. When the user calls `voltage()` or `voltage.get()`, the call is redirected to that function-object. It's the same as with function `_set_voltage` which is passed as argument `set_cmd`.

The only difference is that `_get_voltage` is only supported by firmware version 4. With firmware version 3 there is no getter for voltage because the instrument doesn't support it. So, `voltage_get` (= `get_cmd`) was set to `None`, so that Qcodes returns the cached value (last setpoint), when calling the getter.

FarBo: My bad. It's right.

qcodes_contrib_drivers/drivers/Attocube/ANC350.py
```diff
@@ -0,0 +1,513 @@
...
+            raise NotImplementedError("Only version 3 and 4 of ANC350's driver-DLL are currently "
+                                      "supported")
+
+        self._lib = library
+        self._device_no = inst_no
+        self._device_handle = self._lib.connect(inst_no)
+
+        axischannels = ChannelList(self, "Anc350Axis", Anc350Axis)
+        for nr, axis in enumerate(['x', 'y', 'z']):
+            axis_name = "{}_axis".format(axis)
```
FarBo, line 482: This formatting is also preferred here:
f"{axis}_axis"

qcodes_contrib_drivers/drivers/Attocube/ANC350Lib/interface.py
```diff
@@ -0,0 +1,158 @@
...
+        Args:
+            code: Occurred error code
+            message: Error message
+
+        Returns:
+            Detailed error message
+        """
+        message_out = "Error in ANC350Lib"
+        if code is not None:
+            message_out += " ({}): ".format(code)
```
FarBo, line 75: Since you changed string format as f"" everywhere, you may change it here

lankes-fzj: Oh, sorry I forgot to check the other files

FarBo: No worries.

FarBo:
@jenshnielsen 
Could you please have a quick review on this PR and approve it if it looks fine?

jenshnielsen (approved):

lankes-fzj:
@FarBo @jenshnielsen Thanks for your support :)

PR #75, opened 2020-09-18
Driver for BlueFors fridges

edumur:
This driver allow the readout of basic information, pressures and temperature, for BlueFors fridges.

### DIFF
--- a/qcodes_contrib_drivers/drivers/BlueFors/BlueFors.py
+++ b/qcodes_contrib_drivers/drivers/BlueFors/BlueFors.py
@@ -1,0 +1,200 @@
+# This Python file uses the following encoding: utf-8
+# Etienne Dumur <etienne.dumur@gmail.com>, september 2020
+
+import os
+import pandas as pd
+from datetime import date
+from qcodes.instrument.base import Instrument
+
+class BlueFors(Instrument):
+    """
+    This is the QCoDeS python driver to extract the temperature and pressure
+    from a BlueFors fridge
+    """
+
+    def __init__(self, name                      : str,
+                       folder_path               : str,
+                       channel_vacuum_can        : int=None,
+                       channel_pumping_line      : int=None,
+                       channel_compressor_outlet : int=None,
+                       channel_compressor_inlet  : int=None,
+                       channel_mixture_tank      : int=None,
+                       channel_venting_line      : int=None,
+                       channel_50k_plate         : int=None,
+                       channel_4k_plate          : int=None,
+                       channel_magnet            : int=None,
+                       channel_still             : int=None,
+                       channel_mixing_chamber    : int=None,
+                       **kwargs) -> None:
+        """
+        QCoDeS driver for BlueFors fridges.
+        When a channel is set to None, it will not appear has an available 
+        parameter, this is usefull for fridge not having magnet.
+
+        Args:
+        name (str): Name of the instrument.
+        folder_path (str): Valid path toward the BlueFors log folder.
+        channel_vacuum_can (int, optional): channel of the vacuum can. Defaults to None.
+        channel_pumping_line (int, optional): channel of the pumping line. Defaults to None.
+        channel_compressor_outlet (int, optional): channel of the compressor outlet. Defaults to None.
+        channel_compressor_inlet (int, optional): channel of the compressor inlet. Defaults to None.
+        channel_mixture_tank (int, optional): channel of the mixture tank. Defaults to None.
+        channel_venting_line (int, optional): channel of the venting line. Defaults to None.
+        channel_50k_plate (int, optional): channel of the 50k plate. Defaults to None.
+        channel_4k_plate (int, optional): channel of the 4k plate. Defaults to None.
+        channel_magnet (int, optional): channel of the magnet. Defaults to None.
+        channel_still (int, optional): channel of the still. Defaults to None.
+        channel_mixing_chamber (int, optional): channel of the mixing chamber. Defaults to None.
+        """
+
+        super().__init__(name = name, **kwargs)
+        
+        self.folder_path = os.path.abspath(folder_path)
+
+        if channel_vacuum_can is not None:
+            self.add_parameter(name       = 'pressure_vacuum_can',
+                               unit       = 'mBar',
+                               get_parser = float,
+                               get_cmd    = lambda: self.get_pressure(channel_vacuum_can),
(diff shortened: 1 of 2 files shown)


edumur:
Dear all,

I would like to make another pull request to propose another driver.
However I can't do so while this pull request is pending.
Is it possible to have a review of this driver or to give me an estimation of the time it will take?

Thanks.

astafan8:
> I would like to make another pull request to propose another driver.
However I can't do so while this pull request is pending.

@edumur why is that? is it also a blue fors driver hence the folder structure is needed?

> Is it possible to have a review of this driver or to give me an estimation of the time it will take?

unfortunately, it's difficult to provide an estimation. yet we'll will try to make the review within two weeks.

-------------

@edumur additional question about the approach to the driver - reading bluefors log files. apparently, it is possible for some of the parameters to just read them from the instrument. i'll post here some information from my colleague who went a bit further in controlling the fridge:

> I already started looking in to being able to control the fridge and reading the status directly while being at it. For the valve control, this is possible via a telnet server (see attached image), latest ValveControl software is required to have access to all parameters (4.4.1). I have also asked Bluefors for their planning on releasing their new software to control the fridge, which will feature an API but it doesn't look like this will be released in the near or not-so near future. For the lakeshore reader (temperatures) this is not possible, but it is possible to connect directly to the instrument itself by multiplying the com-port of the instrument, where the instances reading/controlling the instrument will talk to the instrument via virtual com-ports. The communication is managed in sequence to the instrument.

telnet

is this of any help/use?

qcodes_contrib_drivers/drivers/BlueFors/BlueFors.py
```diff
@@ -0,0 +1,200 @@
...
+        
+        folder_name = date.today().strftime("%y-%m-%d")
+        file_path = os.path.join(self.folder_path, folder_name, 'CH'+str(channel)+' T '+folder_name+'.log')
+        
+        df = pd.read_csv(file_path,
+                            delimiter = ',',
+                            names     = ['date', 'time', 'y'],
+                            header    = None)
+
+        # There is a space before the day
```
astafan8, line 165: is this also true for two-digit days?

qcodes_contrib_drivers/drivers/BlueFors/BlueFors.py
```diff
@@ -0,0 +1,200 @@
...
+        return df.iloc[-1]['y']
+
+
+    def get_pressure(self, channel: int) -> float:
+        """
+        Return the last registered pressure of the current day for the
+        channel.
+        
+        Args:
+            channel (int): Channel from which the pressure is extracted.
```
astafan8, line 177: there's no need to specify the type in the docstring if it's in the function signature
```suggestion
            channel: Channel from which the pressure is extracted.
```

qcodes_contrib_drivers/drivers/BlueFors/BlueFors.py
```diff
@@ -0,0 +1,200 @@
...
+                       channel_pumping_line      : int=None,
+                       channel_compressor_outlet : int=None,
+                       channel_compressor_inlet  : int=None,
+                       channel_mixture_tank      : int=None,
+                       channel_venting_line      : int=None,
+                       channel_50k_plate         : int=None,
+                       channel_4k_plate          : int=None,
+                       channel_magnet            : int=None,
+                       channel_still             : int=None,
+                       channel_mixing_chamber    : int=None,
```
astafan8, line 27: these can't be None by default because if they are None then the get_pressure method will not work, right? so either the driver should to NOT create (or expose) the pressure parameters for those channels or get_pressure method should be made smarter to raise a meaninful exception if None is being passed to it.

qcodes_contrib_drivers/drivers/BlueFors/BlueFors.py
edumur: Yes:

qcodes_contrib_drivers/drivers/BlueFors/BlueFors.py
edumur: My bad, I am using the "python docstring generator" on vscode and the type was added automatically...

qcodes_contrib_drivers/drivers/BlueFors/BlueFors.py
edumur: Hum I am not sure I am following your comment.
However seeing the driver now after some time I recognize that this is weird.
The main purpose was to allow a driver to being instanced without magnet, which is an option.
I could put all channel mandatory and only the magnet as optional, what do you think?

edumur:
The new driver is about an triton oxford fridge, it is not related to the bluefors pull request.

Thank you for your information about reading some parameters directly from the fridge.
The idea of the driver is the following, it should be easy and fast and currently I do not know how to do that for any fridges of my lab.
Reading temperature, usually, takes a very long time and since I want to record the temperature of the fridge in the "metadata" for all my experiments I am using the last recorded log temperature which is correct enough in a steady state and super fast.
This approach has the clear advantage to be easy to implement but can't be used to control the fridge.

qcodes_contrib_drivers/drivers/BlueFors/BlueFors.py
astafan8: sorry, this is my bad, i misread the `if channel_magnet is not None` before every add_parameter call, all is good.

> The main purpose was to allow a driver to being instanced without magnet, which is an option.
I could put all channel mandatory and only the magnet as optional, what do you think?

if it makes sense to also use the driver without mixing_chamber and 4k_plate, etc etc - then yes. if it only makes sense to use the driver with/without magnet ONLY, then only the magnet channel should be optional.

astafan8:
then the following is worth being mentioned in the driver's docstring:

> Thank you for your information about reading some parameters directly from the fridge.
The idea of the driver is the following, it should be easy and fast and currently I do not know how to do that for any fridges of my lab.
Reading temperature, usually, takes a very long time and since I want to record the temperature of the fridge in the "metadata" for all my experiments I am using the last recorded log temperature which is correct enough in a steady state and super fast.
This approach has the clear advantage to be easy to implement but can't be used to control the fridge.

edumur:
I believed to have implemented everything, please tell me otherwise.

astafan8 (approved):
thanks!

PR #79, opened 2020-10-12
Driver for Oxford Triton fridge 

edumur:
This driver allow the readout of basic information, pressures and temperature, for the Oxford Triton fridges.
As written in the driver, these information are gotten from the fridge log files which are themselves converted from vcl to csv format.


edumur:
@Akshita07 Are you using some kind of code formatter like black?

Akshita07:
@edumur Use of Black is not mandatory but it's a great tool. We do follow python code writing best practices in accordance to PEP 8.

qcodes_contrib_drivers/drivers/Oxford/Triton.py
```diff
@@ -0,0 +1,201 @@
...
+                           get_cmd=lambda: self.get_temperature('mc'),
+                           docstring='Temperature of the mixing chamber',)
+        
+        self.connect_message()
+
+    def vcl2csv(self) -> str:
+        """
+        Convert vcl file into csv file using proprietary binary exe.
+        The executable is called through the python subprocess library.
+        To avoid to frequent file conversion, a timer of self._conversion_time
```
astafan8, line 115: typo?
```suggestion
        To avoid to frequent file conversion, a timer of self.conversion_timer
```

Akshita07:
@edumur The file VCL_2_ASCII_CONVERTER.exe that you have added for this driver, has it been downloaded from oxford's website? If that is the case then you do not need to add it into the driver, instead document where and how to take it from. If you found it by digging into the oxford software then that needs to be documented in the driver for clarity.

qcodes_contrib_drivers/drivers/Oxford/Triton.py
edumur: Yes.
It is corrected now, thank you!

edumur:
@Akshita07 I clarified the docstring of the class to make the origin of the VCL_2_ASCII_CONVERTER file more obvious.

qcodes_contrib_drivers/drivers/Oxford/Triton.py
```diff
@@ -0,0 +1,204 @@
...
+        The executable is called through the python subprocess library.
+        To avoid to frequent file conversion, a timer of self.conversion_timer
+        second is used.
+
+        Returns:
+            str: The output of the bash command
+        """
+        
+        conversion = False
+        if self._timer+self.rr <= time.time():
```
Akshita07, line 126: Check if self.rr is correct.

Akshita07:
@edumur Since VCL_2_ASCII_CONVERTER is provided with Oxford Instrument. It can be removed from the driver. And you can update the driver to remove hard coded path and add checks which ensure that this file's path is provided during instantiation of the driver.

qcodes_contrib_drivers/drivers/Oxford/Triton.py
edumur: I have no idea how this happened, corrected now.

edumur:
@Akshita07 Hi, I ran some tests and the assert condition you added made the driver bugged. I so removed it. If the assert served a purposed, please tell me so that I can add it again while making it works.

I am not sure to follow your idea about the VCL_2_ASCII_CONVERTER file. Currently it is working, why changing thing?
Also the converter path is an optional parameter of the class and it is not hard coded, did I missed something?

Thanks again!

Akshita07:
Hi @edumur ,

The `assert False` statement was added by me because mypy fails without it on Triton.py file and if mypy fails then this driver can not be accepted for merge. If you want to remove this statement please add an appropriate `else` statement for `vcl2csv` function for `if` statement at line 130 and make sure that mypy passes. 

The reason behind my advice regarding  VCL_2_ASCII_CONVERTER.exe is - it is not preferred to add extra files along with the driver which are available readily with the instrument. 

You are seeing converter_path parameter as an optional parameter because you have hard coded the converter_path parameter's default value in line 19 of Triton.py as `converter_path: str = 'VCL_2_ASCII_CONVERTER.exe'`. If you remove this binary from the driver folder, which I suggest you remove, then you cannot set this default path value and your converter_path will not be optional anymore. You have to provide this file's path while instantiation of the driver and which will be used whenever `conversion` is true in line 130 of Triton.py. 

I hope it helps in improving your driver.

edumur:
Hi @Akshita07,

It was very clear thank you!
I updated the driver.

docs/examples/OxfordTriton.ipynb
```diff
@@ -0,0 +1,149 @@
...
+   "pygments_lexer": "ipython3",
+   "version": "3.7.7"
+  },
+  "nbsphinx": {
+   "execute": "never"
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 4
+}
```
Akshita07, line 149: Please add 1 empty line at the end of code in this file.

edumur: Done.

qcodes_contrib_drivers/drivers/Oxford/Triton.py
```diff
@@ -0,0 +1,213 @@
...
+        second is used.
+
+        Returns:
+            str: The output of the bash command
+        """
+        
+        conversion = False
+        if self._timer+self.conversion_timer <= time.time():
+            conversion = True
+        elif not os.path.isfile(self.file_path[:-3]+'txt'):
```
Akshita07, line 136: The expression will never be true as it has already been checked if given file_path is correct in line 54.

edumur: The test line 35 tests if a vcl file path is correctly given.
That vcl file is the original oxford log file.

The test line 135 tests if a txt file with the same file path as the vcl file is present.
That txt file is created by the driver through the binary vcl converter.
That condition is true the first time the function is called, when the timer hasn't reached is threshold but no txt file already exist.

Does that make sense?

Akshita07: Sorry, I should have been more clear. I meant expression in line 136 `not os.path.isfile(self.file_path[:-3]+'txt')`. You have already checked for it in line 54.

edumur: `file_path` is the path towards the **vcl** file.

`not os.path.isfile(self.file_path[:-3]+'txt')` check the presence of the converted **txt** file.

Akshita07:
@astafan8 Please have a quick look and approve the PR if you see fit.

astafan8 (approved):

PR #82, opened 2020-11-09
Adds notebook example for SQ

Tjalling-a-a:
See title :)


Tjalling-a-a:
@jenshnielsen 
I wanted to add this example, but the pipeline times out (because the device obviously can't be reached).
Any tips on how to fix this?

jenshnielsen:
@Tjalling-a-a You can mark the notebook as non executable in its metadata. See here https://qcodes.github.io/Qcodes/examples/writing_drivers/Creating-Instrument-Drivers.html?highlight=execute#Documentation and the link within

PR #85, opened 2020-12-08
Feature/keithley 2000 scan

MarceloSemO:
New features:
- added class for Keithley 2000-SCAN card
- added scanning card to Keithely_6500 class
- Keithley 2000-SCAN can be included into other DMM classes if necessary
- Keithley DMM6500: Check if correct terminal is activated
- Keithley DMM6500:  Allow for simpler access of parameters (e.g. `dmm.resistance.get()` instead of `dmm.measure.res.get()`. The latter still works to ensure "downwards" compatibilty.)

### DIFF
--- a/docs/examples/Tektronix_Keithley_6500.ipynb
+++ b/docs/examples/Tektronix_Keithley_6500.ipynb
@@ -40,20 +40,20 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 3,
    "metadata": {},
    "outputs": [
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Connected to: KEITHLEY INSTRUMENTS DMM6500 (serial:04475465, firmware:1.0.04b) in 0.06s\n",
+      "Connected to: KEITHLEY INSTRUMENTS DMM6500 (serial:04438044, firmware:1.0.04b) in 0.07s\n",
       "Scanner card 2000-SCAN detected.\n"
      ]
     }
    ],
    "source": [
-    "dmm = dmm6500.Keithley_6500('dmm-1','USB0::0x05E6::0x6500::04475465::INSTR')"
+    "dmm = dmm6500.Keithley_6500('dmm-1','TCPIP0::192.168.1.12::inst0::INSTR')"
    ]
   },
   {
@@ -72,16 +72,16 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": 6,
    "metadata": {},
    "outputs": [
     {
      "data": {
       "text/plain": [
-       "511.8029"
+       "5799.959"
       ]
      },
-     "execution_count": 3,
+     "execution_count": 6,
      "metadata": {},
      "output_type": "execute_result"
     }
@@ -115,16 +115,16 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 4,
+   "execution_count": 7,
    "metadata": {},
    "outputs": [
     {
      "data": {
       "text/plain": [
-       "511.8136"
+       "5793.865"
       ]
      },
-     "execution_count": 4,
+     "execution_count": 7,
      "metadata": {},
      "output_type": "execute_result"
     }
@@ -149,7 +149,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 5,
+   "execution_count": 8,
    "metadata": {},
    "outputs": [
     {
@@ -158,7 +158,7 @@
        "'FRON'"
       ]
      },
-     "execution_count": 5,
+     "execution_count": 8,
      "metadata": {},
      "output_type": "execute_result"
     }
@@ -176,7 +176,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 7,
+   "execution_count": 9,
    "metadata": {},
    "outputs": [
     {
@@ -185,7 +185,7 @@
        "'REAR'"
       ]
      },
-     "execution_count": 7,
+     "execution_count": 9,
      "metadata": {},
      "output_type": "execute_result"
     }
@@ -210,16 +210,16 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 8,
+   "execution_count": 10,
    "metadata": {},
    "outputs": [
     {
      "data": {
       "text/plain": [
-       "998.4387"
+       "5798.519"
       ]
      },
-     "execution_count": 8,
+     "execution_count": 10,
      "metadata": {},
      "output_type": "execute_result"
     }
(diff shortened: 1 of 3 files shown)


FarBo:
@MarceloSemO 

Thank you for the contribution. I will start the review ASAP.

FarBo:
I have kind of general comments from skimming in the files:
1. Please provide a little bit of docstring for classes and functions (ideally, all arguments in the classes and functions could have some descriptions)
2. All defined parameters in both .py files should have validators specifying data type and min/max if required (you can find validators in qcodes.utils.validators)
3. Please provide typing for all arguments

MarceloSemO:
Thanks for the feedback, I added docstrings and typings. Two problems/questions are remaining:

1. Should parameter which are only gettable also have a validator? I am asking because all settable parameters in my drivers already have validators, but you still asked me to add validators to the parameters. As far as I understood, the purpose of a validator is to prevent to send values which could cause unpredictable behavior of the device. In case of non-settable parameters this cannot happen anyways. Please correct me if I am mistaken and if I should also add validators to non-settable parameters.

2. The test now fails, because the autodoc fails to import the module Keithley_2000_Scan. I don't know how to solve this, since I the only import of the Keithley_2000_Scan module happens in the Keithley_6500 module, and I did not change the import statement there. What I changed is that I imported the Keithley_6500 module in the Keithley_2000_Scan module, since I needed this to provide the desired type hint for the variable "dmm" in the __init__ function of the Keithley_2000_Scan_Channel class. My guess is that the circular import is what causes the test to failm but I don't know how to fix, so I would appreciate your help.

jenshnielsen:
@MarceloSemO I pushed a small fix for the circular import. Basically you need to only import the Parent instrument into the scan module when you are typechecking

qcodes_contrib_drivers/drivers/Tektronix/Keithley_6500.py
```diff
@@ -2,55 +2,154 @@
...
 from functools import partial
+from .Keithley_2000_Scan import Keithley_2000_Scan_Channel
 
 
 class Keithley_Sense(InstrumentChannel):
-    def __init__(self, parent: VisaInstrument, name: str, channel: str) -> None:
+    """
+    This is the class for a measurement channel, i.e. the quantity to be measured (e.g. resistance, voltage).
+    """
+    def __init__(self, parent: VisaInstrument, name: str, quantity: str) -> None:
```
jenshnielsen, line 12: It usually not a good idea to rename an variable in a method definition even if the new name is better since this will break the use for any one else using the driver that passes the variable as a keyword argument.

qcodes_contrib_drivers/drivers/Tektronix/Keithley_6500.py
```diff
@@ -2,55 +2,154 @@
...
+        self.add_parameter('nplc',
+                           label='NPLC',
                            get_parser=float,
-                           get_cmd=f"SENS:{channel}:NPLC?",
-                           set_cmd=f"SENS:{channel}:NPLC {{:.4f}}",
-                           vals=Numbers(0.0005, 12)
+                           get_cmd=f"SENS:{quantity}:NPLC?",
+                           set_cmd=f"SENS:{quantity}:NPLC {{:.4f}}",
+                           vals=Numbers(0.0005, 12),
+                           docstring="Integration rate (Numbers Per Line Cycle)"
```
jenshnielsen, line 39: I think NPLC is normally short for ` Number of Power Line Cycles`

qcodes_contrib_drivers/drivers/Tektronix/Keithley_6500.py
```diff
@@ -2,55 +2,154 @@
...
+        scan_idn_msg = self.ask(":SYSTem:CARD1:IDN?")
+        if scan_idn_msg != "Empty Slot":
+            msg_parts = scan_idn_msg.split(",")
+            print(f"Scanner card {msg_parts[0]}-SCAN detected.")
+            for ch_number in range(1, 11):
+                scan_channel = Keithley_2000_Scan_Channel(self, ch_number)
+                self.add_submodule(f"ch{ch_number:d}", scan_channel)
+
+    # only measure if front terminal is active
+    def measure(self, quantity: str) -> str:
```
jenshnielsen, line 141: As far as I can see the user should never call this directly so perhaps it should be called `_measure` to indicate that is

qcodes_contrib_drivers/drivers/Tektronix/Keithley_2000_Scan.py
```diff
@@ -0,0 +1,66 @@
...
+                           get_parser=float,
+                           get_cmd=partial(self.measure, 'VOLT'))
+
+        self.add_parameter('current_dc',
+                           unit='A',
+                           label=f'DC current CH{self.channel}',
+                           get_parser=float,
+                           get_cmd=partial(self.measure, 'CURR'))
+
+    def measure(self, quantity: str) -> str:
```
jenshnielsen, line 50: This should perhaps also be private

qcodes_contrib_drivers/drivers/Tektronix/Keithley_6500.py
```diff
@@ -2,55 +2,154 @@
...
+            quantity: Quantity to be measured
+
+        Returns: Measurement result
+
+        """
+        if self.active_terminal.get() == 'FRON':
+            return self.ask(f"MEAS:{quantity}?")
+        else:
+            print("WARNING: Rear terminal is active instead of front terminal.")
+            return "nan"
```
jenshnielsen, line 155: Perhaps rather than printing and returning nan this should raise an error

```
else:
   Raise RuntimeError("Rear terminal is active instead of front terminal.")
```
?

jenshnielsen (approved):
I agree that validators are not needed for get only parameters. I left a few suggestions inline

FarBo:
@MarceloSemO 

Regarding the first question, there is no need for validators for only-gettable parameters.

Thanks for the changes. I am reviewing it in a continuous basis.

qcodes_contrib_drivers/drivers/Tektronix/Keithley_2000_Scan.py
```diff
@@ -0,0 +1,66 @@
+from functools import partial
+from typing import TYPE_CHECKING
+
+from qcodes.instrument import InstrumentChannel
```
FarBo, line 4: It's better to import from the full path: from qcodes.instrument.channel import InstrumentChannel

qcodes_contrib_drivers/drivers/Tektronix/Keithley_2000_Scan.py
```diff
@@ -0,0 +1,66 @@
...
+        Returns: Measurement result
+
+        """
+        if self.dmm.active_terminal.get() == 'REAR':
+            self.write(f"SENS:FUNC '{quantity}', (@{self.channel:d})")
+            self.write(f"ROUT:CLOS (@{self.channel:d})")
+            return self.ask("READ?")
+        else:
+            print("WARNING: Front terminal is active instead of rear terminal.")
+            return "nan"
```
FarBo, line 66: Similar to Jens comment for Keithley_6500.py, this may also be raised as an error, unless you have a reason for this warning message.
Just a side note: for nan, you can use numpy.nan

qcodes_contrib_drivers/drivers/Tektronix/Keithley_6500.py
```diff
@@ -2,55 +2,154 @@
 from qcodes.instrument import InstrumentChannel
```
FarBo, line 2: It's better to import from the full path: from qcodes.instrument.channel import InstrumentChannel

qcodes_contrib_drivers/drivers/Tektronix/Keithley_6500.py
MarceloSemO: Thank you for the remark, I changed it back to the previous name "channel".

MarceloSemO:
@jenshnielsen and @FarBo, thank you very much for your comments! I have implemented the changes you suggested and performed a rebase with respect to the latest commit on the master branch. I think the pull request should now be ready to be merged.

qcodes_contrib_drivers/drivers/Tektronix/Keithley_2000_Scan.py
```diff
@@ -0,0 +1,65 @@
+from functools import partial
+from typing import TYPE_CHECKING
+
+from qcodes.instrument.channel import InstrumentChannel
```
FarBo, line 4: The previous import was the correct import according to qcodes public API.
I deleted my comment in this regard right after making that comment, but you changed it anyway. So please revert back to the previous import in both files.

docs/examples/Tektronix_Keithley_6500.ipynb
```diff
@@ -74,6 +89,212 @@
...
+   ],
+   "source": [
+    "dmm.ch1.resistance.get()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Checking the active terminal"
```
FarBo, line 235: You probably need to rerun the notebook, because with the new changes, this part should throw error instead of these warning messages.

MarceloSemO:
I have now re-run the notebook and implemented the required changes. I also performed a rebase, now the PR should be ready to be merged.

FarBo (approved):

PR #86, opened 2020-12-14
feat: add driver for Lakeshore Model 625 

simonzihlmann:
This pull request proposes a driver for the Lakeshore Model 625 superconducting magnet power supply as well as an example notebook.


simonzihlmann:
Please let me know why the checks failed and if some work needs to be done. I din't get the errors.

jenshnielsen:
@simonzihlmann

There are two seperate issues. 

1) The CI is trying to execute your notebook. That cannot work since it does not have access to the instrument. You should disable execution as documented here https://qcodes.github.io/Qcodes/examples/writing_drivers/Creating-Instrument-Drivers.html#Documentation

2) Is a typechecking issue. I will leave a comment inline explaining this

qcodes_contrib_drivers/drivers/Lakeshore/Model_625.py
```diff
@@ -0,0 +1,524 @@
...
+
+    This class uses T/A and A/s as units.
+
+    Args:
+        name (str): a name for the instrument
+        address (str): VISA address of the device
+        current_ramp_limit: A current ramp limit, in units of A/s
+    """
+
+    def __init__(self, name: str, coil_constant: float,  field_ramp_rate: float, address: str=None,
```
jenshnielsen, line 23: address cannot be None within the VisaInstrument class. I think you can just remove the `=None` part here to fix it

simonzihlmann: Thanks for pointing out these two problems. I have fixed them and the tests pass now. Let me know if additional work needs to be done.

qcodes_contrib_drivers/drivers/Lakeshore/Model_625.py
```diff
@@ -0,0 +1,524 @@
...
+    
+        # Add reset function
+        self.add_function('reset', call_cmd='*RST')
+        if reset:
+            self.reset()
+
+        # Add power supply parameters
+        self.add_parameter(name='current_limit',
+                           unit="A",
+                           set_cmd=lambda x: self._set_curent_limit(x),
```
jenshnielsen, line 36: I don't think you need the lambda functions here. As far as I can see `set_cmd=self._set_curent_limit` should be equivalent

jenshnielsen: Same in the parameters below

qcodes_contrib_drivers/drivers/Lakeshore/Model_625.py
```diff
@@ -0,0 +1,524 @@
...
+                           get_cmd='RSEG?',
+                           get_parser=float,
+                           val_mapping={'disabled': 0,
+                                        'enabled': 1}
+                           )
+        
+        self.add_parameter(name='persistent_switch_heater',
+                           set_cmd=lambda x: self._set_persistent_switch_heater_status(x),
+                           get_cmd=self._get_persistent_switch_heater_status,
+                           get_parser=float,
```
jenshnielsen, line 95: looks like this should perhaps be `get_parser=int` ? 

Note that qcodes as a create_on_off_val_mapping that you can consider using here https://qcodes.github.io/Qcodes/api/utils/helpers.html?highlight=val_mapping#qcodes.utils.helpers.create_on_off_val_mapping

jenshnielsen (commented):
@simonzihlmann Thanks for the contribution. I left a few minor comments inline

qcodes_contrib_drivers/drivers/Lakeshore/Model_625.py
```diff
@@ -0,0 +1,524 @@
...
+        self.add_parameter(name='coil_constant',
+                           unit = self.coil_constant_unit,
+                           set_cmd=self._update_coil_constant,
+                           get_cmd=self._get_coil_constant,
+                           get_parser=float,
+                           vals=Numbers(0.001, 999.99999)  # what are good numbers here?
+                           )
+        
+        self.add_parameter(name='field',
+                           full_name = self.name + '_' + 'field',
```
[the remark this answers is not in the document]
simonzihlmann: I see, thanks for pointing this out.

qcodes_contrib_drivers/drivers/Lakeshore/Model_625.py
simonzihlmann: I guess so. Don't know why some parameter set functions where like this. I guess I have copied from several other drivers... Thanks for spotting this.

qcodes_contrib_drivers/drivers/Lakeshore/Model_625.py
simonzihlmann: I changed to int. 
Thanks for the hint about the on_off_val_mapping, but since it works like it is, I would leave it.

qcodes_contrib_drivers/drivers/Lakeshore/Model_625.py
```diff
@@ -0,0 +1,523 @@
+import logging
+import time
+from typing import Union, Tuple
+
+
+from qcodes import VisaInstrument
+from qcodes.utils.validators import  Numbers, Enum
+
+log = logging.getLogger(__name__)
```
jenshnielsen, line 9: I think this can safely be removed as `self.log` is used

qcodes_contrib_drivers/drivers/Lakeshore/Model_625.py
```diff
@@ -0,0 +1,523 @@
...
+class Lakeshore625(VisaInstrument):
+    """
+    Driver for the Lakeshore Model 625 superconducting magnet power supply.
+
+    This class uses T/A and A/s as units.
+
+    Args:
+        name (str): a name for the instrument
+        address (str): VISA address of the device
+        current_ramp_limit: A current ramp limit, in units of A/s
```
jenshnielsen, line 20: This seems be called `field_ramp_rate` below

qcodes_contrib_drivers/drivers/Lakeshore/Model_625.py
```diff
@@ -0,0 +1,523 @@
...
+
+    def _get_oer_quench_bit(self) -> int:
+        """
+        Returns the oer quench bit
+
+        Returns
+        -------
+            quench bit
+        """
+        return self._get_operational_errors()[3]
```
jenshnielsen, line 412: Mypy complains since this returns a str but you expect an int. Either convert the output to int before returning or change the signature.

simonzihlmann: Thanks, I saw the error. Took me some time to find the output of mypy. I converted to int now.

qcodes_contrib_drivers/drivers/Lakeshore/Model_625.py
simonzihlmann: True, my bad. One should specify the ramp rate when initializing the instrument.,

qcodes_contrib_drivers/drivers/Lakeshore/Model_625.py
simonzihlmann: Ok, I removed `log = logging.getLogger(__name__)`

simonzihlmann:
@jenshnielsen Do I miss anything? Let me know if there are still problems with the driver.

jenshnielsen:
@simonzihlmann sorry for the delay. I think this looks good

simonzihlmann:
> @simonzihlmann sorry for the delay. I think this looks good

nor worry

jenshnielsen (approved):

PR #89, opened 2021-01-20
Driver for ITest from Bilt

edumur:
Allow set and get of voltage with different ramps.

### DIFF
--- a/docs/examples/ITest.ipynb
+++ b/docs/examples/ITest.ipynb
@@ -75,7 +75,7 @@
     {
      "output_type": "stream",
      "name": "stdout",
-     "text": "Connected to: 2142 \"ITEST BE2142B/12V 15mA DC-SOURCE/SN04-016 LC1846 VL446\\240\" (serial:None, firmware:None) in 3.63s\n"
+     "text": "Connected to: 2142 \"ITEST BE2142B/12V 15mA DC-SOURCE/SN04-016 LC1846 VL446\\240\" (serial:None, firmware:None) in 0.61s\n"
     }
    ],
    "source": [
@@ -93,7 +93,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": 4,
    "metadata": {
     "tags": []
    },
@@ -114,7 +114,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 4,
+   "execution_count": 5,
    "metadata": {
     "tags": []
    },
@@ -122,7 +122,7 @@
     {
      "output_type": "stream",
      "name": "stdout",
-     "text": "channel 1 voltage 9.76e-08 V\nchannel 2 voltage 5.29e-07 V\nchannel 3 voltage 1.45e-08 V\nchannel 4 voltage 3.26e-07 V\nchannel 5 voltage 1.49e-07 V\nchannel 6 voltage -5.21e-07 V\nchannel 7 voltage -9.63e-08 V\nchannel 8 voltage 7.69e-07 V\nchannel 9 voltage 3.86e-07 V\nchannel 10 voltage 1.37e-06 V\nchannel 11 voltage 3.42e-07 V\nchannel 12 voltage -1.11e-06 V\nchannel 13 voltage 9.32e-07 V\nchannel 14 voltage 1.87e-07 V\nchannel 15 voltage 1.43e-07 V\nchannel 16 voltage -2.42e-07 V\n"
+     "text": "channel 1 voltage 9.92e-07 V\nchannel 2 voltage 1.12e-06 V\nchannel 3 voltage 1.45e-08 V\nchannel 4 voltage -8.67e-07 V\nchannel 5 voltage -1.49e-07 V\nchannel 6 voltage -8.19e-07 V\nchannel 7 voltage -6.93e-07 V\nchannel 8 voltage 1.73e-07 V\nchannel 9 voltage 8.77e-08 V\nchannel 10 voltage -1.17e-07 V\nchannel 11 voltage 1.24e-06 V\nchannel 12 voltage -1.11e-06 V\nchannel 13 voltage 3.78e-08 V\nchannel 14 voltage -7.08e-07 V\nchannel 15 voltage 1.04e-06 V\nchannel 16 voltage -2.42e-07 V\n"
     }
    ],
    "source": [
@@ -140,36 +140,27 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [
-    {
-     "output_type": "execute_result",
-     "data": {
-      "text/plain": "None"
-     },
-     "metadata": {},
-     "execution_count": 5
-    }
-   ],
+   "execution_count": 9,
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
    "source": [
     "dac.ch04.v(0.5)"
    ]
   },
   {
    "cell_type": "code",
-   "execution_count": 5,
-   "metadata": {
-    "tags": []
-   },
+   "execution_count": 10,
+   "metadata": {},
    "outputs": [
     {
      "output_type": "execute_result",
      "data": {
-      "text/plain": "0.50000103"
+      "text/plain": "0.50000014"
      },
      "metadata": {},
-     "execution_count": 5
+     "execution_count": 10
     }
    ],
    "source": [
@@ -185,7 +176,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 6,
+   "execution_count": 12,
    "metadata": {
     "tags": []
    },
@@ -193,25 +184,18 @@
     {
      "output_type": "stream",
      "name": "stdout",
-     "text": "0.50000103\n0.47945673\n0.45147469\n0.42345422\n0.39545613\n0.37143639\n0.34345409\n0.31545213\n0.28745614\n0.26345791\n"
+     "text": "0.50000014\n0.47945107\n0.45145621\n0.42344528\n0.39945953\n0.37142712\n0.34345707\n0.31545273\n0.29144462\n0.26346268\n"
     }
    ],
    "source": [
     "dac.ch04.output_mode('ramp')\n",
     "dac.ch04.ramp_slope(0.00025)\n",
-    "dac.ch04.synchronous(False)\n",
+    "dac.ch04.synchronous_enable(False)\n",
     "dac.ch04.v(0.05)\n",
     "for i in range(10):\n",
     "    print(dac.ch04.v())\n",
     "    time.sleep(0.1)"
    ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": []
   }
  ]
 }
--- a/docs/examples/ITest.ipynb
+++ b/docs/examples/ITest.ipynb
@@ -1,0 +1,190 @@
+{
+ "metadata": {
+  "nbsphinx": {
+   "execute": "never"
+  },
+  "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.8.3-final"
+  },
+  "orig_nbformat": 2,
+  "kernelspec": {
+   "name": "Python 3.8.3 64-bit ('base': conda)",
+   "display_name": "Python 3.8.3 64-bit ('base': conda)",
+   "metadata": {
+    "interpreter": {
+     "hash": "04a291d3d47d4de40481f20ec3463187616b5eceabad1549fe24b847b2edd3ef"
+    }
+   }
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 2,
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Example with the iTest from Bilt"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Import"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "import qcodes as qc\n",
+    "from qcodes_contrib_drivers.drivers.Bilt.ITest import ITest"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Load instrument\n",
+    "\n",
+    "Here the import is done by USB"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [
+    {
+     "output_type": "stream",
+     "name": "stdout",
+     "text": "Connected to: 2142 \"ITEST BE2142B/12V 15mA DC-SOURCE/SN04-016 LC1846 VL446\\240\" (serial:None, firmware:None) in 0.39s\n"
+    }
+   ],
+   "source": [
+    "dac = ITest(name='dac',\n",
+    "            address='ASRL4::INSTR',\n",
+    "            init_start=True)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Initiallization"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "for channel in dac.channels:\n",
+    "    channel.v(0)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Measurement"
+   ]
+  },
+  {
+   "source": [
+    "User can loop over the channel"
+   ],
+   "cell_type": "markdown",
+   "metadata": {}
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [
+    {
+     "output_type": "stream",
+     "name": "stdout",
+     "text": "channel 1 voltage 9.92e-07 V\nchannel 2 voltage -9.61e-07 V\nchannel 3 voltage 3.13e-07 V\nchannel 4 voltage 3.26e-07 V\nchannel 5 voltage 5.00e-01 V\nchannel 6 voltage -5.21e-07 V\nchannel 7 voltage -9.91e-07 V\nchannel 8 voltage -1.26e-07 V\nchannel 9 voltage 3.86e-07 V\nchannel 10 voltage 4.79e-07 V\nchannel 11 voltage 3.42e-07 V\nchannel 12 voltage -2.14e-07 V\nchannel 13 voltage 3.36e-07 V\nchannel 14 voltage -1.11e-07 V\nchannel 15 voltage -7.52e-07 V\nchannel 16 voltage -5.40e-07 V\n"
+    }
+   ],
+   "source": [
+    "for channel in dac.channels:\n",
+    "    print('channel {} voltage {:.2e} V'.format(channel.chan_num, channel.v()))"
+   ]
+  },
+  {
+   "source": [
+    "We can set a voltage on a specific channel but without trigger the voltage is not output."
+   ],
+   "cell_type": "markdown",
+   "metadata": {}
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [
+    {
+     "output_type": "execute_result",
+     "data": {
+      "text/plain": "(None, 3.2574099e-07)"
+     },
+     "metadata": {},
+     "execution_count": 5
+    }
+   ],
+   "source": [
+    "dac.ch04.v(0.5), dac.ch04.v()"
+   ]
+  },
+  {
+   "source": [
+    "A shortcut is available to set and trigger a channel."
+   ],
+   "cell_type": "markdown",
+   "metadata": {}
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {},
+   "outputs": [
+    {
+     "output_type": "execute_result",
+     "data": {
+      "text/plain": "(None, 0.50000073)"
+     },
+     "metadata": {},
+     "execution_count": 6
+    }
+   ],
+   "source": [
+    "dac.ch04.v_and_trig(0.05), dac.ch04.v()"
+   ]
+  }
+ ]
+}
--- a/docs/examples/ITest.ipynb
+++ b/docs/examples/ITest.ipynb
@@ -140,6 +140,24 @@
   },
   {
    "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [
+    {
+     "output_type": "execute_result",
+     "data": {
+      "text/plain": "None"
+     },
+     "metadata": {},
+     "execution_count": 5
+    }
+   ],
+   "source": [
+    "dac.ch04.v(0.5)"
+   ]
+  },
+  {
+   "cell_type": "code",
    "execution_count": 5,
    "metadata": {
     "tags": []
@@ -148,14 +166,14 @@
     {
      "output_type": "execute_result",
      "data": {
-      "text/plain": "(None, 0.50000103)"
+      "text/plain": "0.50000103"
      },
      "metadata": {},
      "execution_count": 5
     }
    ],
    "source": [
-    "dac.ch04.v(0.5), dac.ch04.v()"
+    "dac.ch04.v()"
    ]
   },
   {
--- a/docs/examples/ITest.ipynb
+++ b/docs/examples/ITest.ipynb
@@ -51,6 +51,7 @@
    },
    "outputs": [],
    "source": [
+    "import time\n",
     "import qcodes as qc\n",
     "from qcodes_contrib_drivers.drivers.Bilt.ITest import ITest"
    ]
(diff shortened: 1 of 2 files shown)


FarBo:
Hi @edumur 
Thank you for the contribution. 
I will start the review process in a continuous basis.
I will look into why the PR is not passing the CI.

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,437 @@
...
+        self.add_parameter('v',
+                           label='Channel {} voltage'.format(chan_num),
+                           unit='V',
+                           docstring='Voltage of the channel in volt.',
+                           get_cmd=partial(self._parent._get_voltage, chan_num),
+                           set_cmd=partial(self._parent._set_voltage, chan_num),
+                           get_parser=float
+                           )
+        
+        self.add_parameter('v_and_trig',
```
astafan8, line 45: a curiosity/design question - why is it such that the `v` parameter that has the more user-friendly name (similary to other qcodes drivers with `volt` and `voltage` parameters) does NOT actually set the voltage to that value? i'd expect that the v parameter DOES set the voltage and there's some target_v parameter that allows to tell the instrument about one's intent to set the voltage to some value.

another remark - once a voltage is being set via v_and_trig parameter, shouldn't parameter v be updated to have an invalid cache (see `.cache.invalidate` for more info), or even set to some neautral value?

edumur: Hi,

Thanks for the remarks.
The trigger story is relative to how the ITest is built.
1. you set a ramp mode.
2. you set a ramp speed.
3. you set a target voltage.
4. you trigger the ramping.

However I discussed with a colleague and he had input that I am going to integrate and come back with a new version.
I will also change the naming to target_voltage which is maybe more meaningful.

edumur:
Hi all,

I remove the trigger thing so now there is only a simple voltage parameter.
I added a fake parameter called "synchronous" which block or not the communication with the instrument.
I updated the example accordingly.

edumur:
Could someone review this driver?
I would like to propose another one after this one?

Thanks!

FarBo:
Hi edmur

Thanks for the reminder. I will review it this week.

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,467 @@
...
+        Args:
+            chan: The 1-indexed channel number
+            synchronous: If the channel is synchronous.
+        """
+        
+        chan_id = self.chan_to_id(chan)
+        self._synchronous[chan_id] = synchronous
+
+
+    def _get_chan_synchronous(self, chan: int) -> None:
```
FarBo, line 457: You are returning self._synchronous[chan_id], but saying return is None.  Please fix ths

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,467 @@
...
+
+        Args:
+            chan: Channel to be named
+            name: Name of the channel
+        """
+        chan_id = self.chan_to_id(chan)
+        self.write(chan_id + 'chan:name "{}"'.format(name))
+
+
+    def chan_to_ic(self, chan:int) -> Tuple[int]:
```
FarBo, line 422: What are the types of 'i' and 'c'?  I think 'i' is a float and 'c' is an int. If yes, then the return type should be Tuple[float, int]

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,467 @@
...
+        Returns:
+            state: Power state 
+        """
+        chan_id = self.chan_to_id(chan)
+        state = self.ask(chan_id + 'OUTP ?')
+        
+        if state == '1':
+            return True
+        elif state == '0':
+            return False
```
FarBo, line 389: What do you expect if state is not '1' or '0'? can they be other than these two options?

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,467 @@
...
+        
+        Args:
+            chan The 1-indexed channel number
+            slope Slope of chXX in V/ms
+        """
+        chan_id = self.chan_to_id(chan)
+        self.write('{}VOLT:SLOP {:.8f}'.format(chan_id, slope))
+
+
+    def _get_ramp_slope(self, chan:int) -> float:
```
FarBo, line 261: The return shoudln't be str?

FarBo:
I put some comments to solve the CI issues. Let's first correct the PR to pass the CI, then I will review the content.

qcodes_contrib_drivers/drivers/Bilt/ITest.py
edumur: Corrected.

qcodes_contrib_drivers/drivers/Bilt/ITest.py
edumur: Corrected.

qcodes_contrib_drivers/drivers/Bilt/ITest.py
edumur: I do not expect `state` to be anything different.
However I added an `else` statement with a error in case of.

qcodes_contrib_drivers/drivers/Bilt/ITest.py
edumur: Yes indeed !

edumur:
Done, thanks for your input!

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,469 @@
...
+        i, c = round(chan_num/4.)+1, chan_num%4
+        self.chan_id = 'i{};c{};'.format(i,c)
+        
+        self.add_parameter('v',
+                           label='Channel {} voltage'.format(chan_num),
+                           unit='V',
+                           docstring='Voltage of the channel in volt.',
+                           get_cmd=partial(self._parent._get_voltage, chan_num),
+                           set_cmd=partial(self._parent._set_voltage, chan_num),
+                           get_parser=float
```
FarBo, line 43: Doesn't it need any validator for set values?
It can prevent setting out of limit values in software level.

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,469 @@
...
+                           )
+        
+        self.add_parameter('ramp_slope',
+                           label='Channel {} ramp slope'.format(chan_num),
+                           unit='V/ms',
+                           docstring='Slope of the ramp in V/ms.',
+                           get_cmd=partial(self._parent._get_ramp_slope, chan_num),
+                           get_parser=float,
+                           set_cmd=partial(self._parent._set_ramp_slope, chan_num),
+                           set_parser=float
```
FarBo, line 61: validators here?

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,469 @@
...
+                           get_parser=str
+                           )
+        
+        self.add_parameter('range',
+                           label = 'Channel {} voltage range'.format(chan_num),
+                           docstring='Range of the channelin volt.',
+                           set_cmd=partial(self._parent._set_chan_range, chan_num),
+                           set_parser=float,
+                           get_cmd=partial(self._parent._get_chan_range, chan_num),
+                           get_parser=float)
```
FarBo, line 78: validators here?

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,469 @@
...
+        
+        self.add_parameter('output_mode',
+                           label='Channel {} output mode'.format(chan_num),
+                           docstring='Mode of the output {exp, ramp}.',
+                           get_cmd=partial(self._parent._get_output_function, chan_num),
+                           set_cmd=partial(self._parent._set_output_function, chan_num),
+                           get_parser=str
+                           )
+        
+        self.add_parameter('range',
```
FarBo, line 72: maybe changing the name to v_range or vrange makes more sense?

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,469 @@
...
+    """
+    
+    def __init__(self,name:str,
+                      address:str,
+                      num_chans:int=16,
+                      init_start:bool=False,
+                      synchronous:bool=True,
+                      synchronous_delay:float=1,
+                      synchronous_threshold:float=1e-5,
+                      **kwargs) -> None:
```
FarBo, line 140: **kwargs: Any

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,469 @@
+# This Python file uses the following encoding: utf-8
+# Loick Le Guevel, 2019
+# Etienne Dumur <etienne.dumur@gmail.com>, 2021
+
+from typing import Union, Tuple
```
FarBo, line 5: please add Any here

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,469 @@
...
+                "synchronous_threshold".
+            synchronous_delay: Time between to voltage measurement in second.
+            synchronous_threshold: Threshold to unblock communication in volt.
+            
+        Returns:
+            ITest object
+        """
+        super().__init__(name, address=address,
+                               terminator='\n',
+                               device_clear=False)
```
FarBo, line 163: You can add **kwargs to this super.__init__ as well?

FarBo:
Thank you for the fix.
I also made some comments in the code. Also, I have a question:
Can pos and neg states be set independently? I ask this because I don't know about the instrument, but I am thinking that setting positive state should disable negative state and vice versa. What do you think?

edumur:
I added your correction.

Concerning your question, I tested both pos and neg request on the device directly and setting one does not disable the other.

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,475 @@
...
+        self.chan_id = 'i{};c{};'.format(i,c)
+        
+        self.add_parameter('v',
+                           label='Channel {} voltage'.format(chan_num),
+                           unit='V',
+                           docstring='Voltage of the channel in volt.',
+                           get_cmd=partial(self._parent._get_voltage, chan_num),
+                           get_parser=float,
+                           set_cmd=partial(self._parent._set_voltage, chan_num),
+                           set_parser=vals.Numbers(-50, 50)
```
FarBo, line 45: You can simply say: vals=vals.Numbers(-50, 50)

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,475 @@
...
+                           )
+        
+        self.add_parameter('ramp_slope',
+                           label='Channel {} ramp slope'.format(chan_num),
+                           unit='V/ms',
+                           docstring='Slope of the ramp in V/ms.',
+                           get_cmd=partial(self._parent._get_ramp_slope, chan_num),
+                           get_parser=float,
+                           set_cmd=partial(self._parent._set_ramp_slope, chan_num),
+                           set_parser=vals.Numbers(0, 1)
```
FarBo, line 63: same here

qcodes_contrib_drivers/drivers/Bilt/ITest.py
FarBo: Are you sure no validators needed here?

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,475 @@
...
+    
+        self.add_parameter('bilt_name',
+                           set_cmd=partial(self._parent._set_chan_name, chan_num),
+                           set_parser=str)
+
+        self.add_parameter('synchronous',
+                           get_cmd=partial(self._parent._get_chan_synchronous, chan_num),
+                           get_parser=bool,
+                           set_cmd=partial(self._parent._set_chan_synchronous, chan_num),
+                           set_parser=bool)
```
FarBo, line 110: Are you sure no validators needed here?

docs/examples/ITest.ipynb
```diff
@@ -0,0 +1,199 @@
...
+     "output_type": "execute_result",
+     "data": {
+      "text/plain": "(None, 0.50000103)"
+     },
+     "metadata": {},
+     "execution_count": 5
+    }
+   ],
+   "source": [
+    "dac.ch04.v(0.5), dac.ch04.v()"
```
FarBo, line 158: Perhaps it's better to split each command in one cell, because in this way, it returns None for dac.ch04.v(0.5), which is misleading.

edumur:
Done!

FarBo:
@edumur 
Thank you for the commits. Could you mark the comments as resolved?

@astafan8 
Could you have a final look?

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,475 @@
...
+                           vals=vals.Numbers(0, 1)
+                           )
+        
+        self.add_parameter('output_mode',
+                           label='Channel {} output mode'.format(chan_num),
+                           docstring='Mode of the output {exp, ramp}.',
+                           get_cmd=partial(self._parent._get_output_function, chan_num),
+                           get_parser=str,
+                           set_cmd=partial(self._parent._set_output_function, chan_num),
+                           set_parser=str
```
astafan8, line 72: is it true that the output_mode can only have a certain strings as values? for example, only "ramp" and "non-ramp". If so, then it will be very useful for users and for correctness to use an Enum validator with those strings:
```suggestion
                           set_parser=str,
                           vals=vals.Enum("ramp", "non-ramp")
```

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,475 @@
...
+                           set_parser=float,
+                           get_cmd=partial(self._parent._get_chan_range, chan_num),
+                           get_parser=float)
+        
+        self.add_parameter('state',
+                           docstring='State of the channel {on, off}.',
+                           get_cmd=partial(self._parent._get_chan_state, chan_num),
+                           get_parser=bool,
+                           set_cmd=partial(self._parent._set_chan_state, chan_num),
+                           vals=vals.OnOff())
```
astafan8, line 88: as per this section in the docs https://qcodes.github.io/Qcodes/examples/writing_drivers/Creating-Instrument-Drivers.html#On/Off-parameters , i highly recommend to implement this parameter as suggested below and adjust `_get_chan_state`/`_set_chan_state` methods to accept "on"/"off" as input (and not perform the conversion to them from a boolean):
```suggestion
        self.add_parameter('enabled',
                           docstring='Whether the channel is enabled or not',
                           get_cmd=partial(self._parent._get_chan_state, chan_num),
                           set_cmd=partial(self._parent._set_chan_state, chan_num),
                           val_mapping=create_on_off_val_mapping(on_val="on", off_val="off")
```

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,475 @@
...
+                           vals=vals.Bool())
+        
+        
+        self.add_function('start',
+                          call_cmd=lambda: self._parent._set_chan_state(self.chan_num, 1))
+        
+        self.add_function('stop',
+                          call_cmd=lambda: self._parent._set_chan_state(self.chan_num, 0))
+    
+        self.bilt_name(f'Chan{chan_num:02d}')
```
astafan8, line 119: this can be included into the definition of the bilt_name parameter with `initial_value=f'Chan{chan_num:02d}'` argument

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,475 @@
...
+                           get_cmd=partial(self._parent._get_chan_synchronous, chan_num),
+                           get_parser=bool,
+                           set_cmd=partial(self._parent._set_chan_synchronous, chan_num),
+                           vals=vals.Bool())
+        
+        
+        self.add_function('start',
+                          call_cmd=lambda: self._parent._set_chan_state(self.chan_num, 1))
+        
+        self.add_function('stop',
```
astafan8, line 116: please refrain from using functions unless it's really needed, see here https://qcodes.github.io/Qcodes/examples/writing_drivers/Creating-Instrument-Drivers.html#Functions, likely simple bound methods would do the job

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,475 @@
...
+                           set_cmd=partial(self._parent._set_ramp_slope, chan_num),
+                           vals=vals.Numbers(0, 1)
+                           )
+        
+        self.add_parameter('output_mode',
+                           label='Channel {} output mode'.format(chan_num),
+                           docstring='Mode of the output {exp, ramp}.',
+                           get_cmd=partial(self._parent._get_output_function, chan_num),
+                           get_parser=str,
+                           set_cmd=partial(self._parent._set_output_function, chan_num),
```
astafan8, line 71: so far as i see the implementation of the _set_output_function method, apparently output_mode can be also supplies as a float or an int? if that is so, then this parameter should allow that, right? or should that feature be removed from _set_output_function ? if the parameter needs to be able to not only accept a few selected strings but selected numbers, then those numbers can also be added to the `vals.Enum`. or if the parameter is supposed to accept a range of numbers then the `vals.MultiType` validator can be used: `vals.MultiType(vals.Enum("ramp", "foo"), vals.Numbers(0, 2.5))`

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,475 @@
...
+
+class iTestMultiChannelParameter(MultiChannelInstrumentParameter):
+    """
+    """
+    
+    def __init__(self, channels, param_name, *args, **kwargs):
+        super().__init__(channels, param_name, *args, **kwargs)
+
+    def trigger(self,*args,**kwargs):
+        print(args,kwargs)
```
astafan8, line 130: is this really needed? :) having a trigger method that just prints its arguments?)) if it's not needed, it's better to remove in order not to confuse the future readers of this code

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,475 @@
...
+        """
+        super().__init__(name, address=address,
+                               terminator='\n',
+                               device_clear=False,
+                               **kwargs)
+        
+        self.idn = self.get_idn()
+        self.num_chans = num_chans
+        self.chan_range = range(1,self.num_chans+1)
+        # Fake parameter allowing to control the synchronicity of every channels
```
astafan8, line 174: is there a reason for these being init-only private attributes as opposed to proper parameters that users can change when using the instrument? just in case, a reminder, parameters can have set_cmd=None and get_cmd=None which makes them "in-memory" parameters, so they don't talk to the instrument, but still maintain a value. and then instead of maintaining the `self._synchronous` dictionary on the instrument level, this information can be conveniently maintained on the channel level

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,475 @@
...
+            mode
+        """
+        chan_id = self.chan_to_id(chan)
+        mode = int(self.ask(chan_id + 'TRIG:INPUT?'))
+        if mode == 0:
+            return 'exp'
+        elif mode == 1:
+            return 'ramp'
+        else:
+            return 'ERROR'
```
astafan8, line 320: shall this raise a proper exception with useful error message? :) otherwise it's less likely that users will understand what is wrong :) something like, `raise ValueError(f"Got unexpected output function mode {mode}")`

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -0,0 +1,475 @@
...
+        if isinstance(pos_sat,(int,float)):
+            self.write(chan_id + 'VOLT:SAT:POS {:.8f}'.format(pos_sat))
+        elif isinstance(pos_sat,str):
+            self.write(chan_id + 'VOLT:SAT:POS MAX')
+
+
+    def _set_chan_neg_sat(self, chan:int,
+                                neg_sat: Union[int, str]) -> None:
+        chan_id = self.chan_to_id(chan)
+        if isinstance(neg_sat,(int,float)):
```
astafan8, line 363: whenever a function accepts a union of numbers and strings, the types should be changed to `Union[float, str]` becasue floats are allowed as well (right?)

astafan8 (approved):
I left some comments, but only a few of them are critical (for example, having the correct types in methods signatures); it's up to you @edumur and @FarBo to judge what's worth changing and what can wait for the future

qcodes_contrib_drivers/drivers/Bilt/ITest.py
FarBo: This is recommended.

FarBo:
Thanks Mikhail, I think most of your comments are recommended and it's up to @edumur how he wants to implement them.

@edumur, if you like to consider these comments, please implement them and I will have a final review and approve it. Then we can merge it. Let me know what you prefer.

edumur:
@FarBo I will work on the nice comments of @astafan8 ;) .

qcodes_contrib_drivers/drivers/Bilt/ITest.py
edumur: Hum, could point.
I could change it and make in an 'in-memory' thing at the channel level but then how would I access it at the instrument level, for example here:
```python
    def _set_voltage(self, chan:int,
                           v_set:float) -> None:
        """
        Set cmd for the chXX_v parameter
        
        Args:
            chan: The 1-indexed channel number
            v_set: The target voltage
        """
        chan_id = self.chan_to_id(chan)
        self.write('{}VOLT {:.8f}'.format(chan_id, v_set))
        self.write(chan_id + 'TRIG:INPUT:INIT')
        if self._get_chan_synchronous(chan):
            v = self._get_voltage(chan)
            while abs(v_set - v)>=self._synchronous_threshold:
                sleep(self._synchronous_delay)
                v = self._get_voltage(chan)
```

Thanks.

astafan8: since the instrument is aware of all the channels and they can be accessed by their name(?), then `self.<chanX>.synchronous_delay()` will just work.

now that i look at this method, it seems to me that it actually also should go inside the channel because there is nothing in this method that depends on the instrument itself. the write/ask method calls on the channel level forward to those on the instrument level, so that you don't have to worry about

edumur:
I, normally, integrated all the comments 🤞.

FarBo:
@edumur 
Could you resolve the comments that are not open?

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -101,22 +107,47 @@ def __init__(self, parent: Instrument,
     
         self.add_parameter('bilt_name',
                            set_cmd=partial(self._parent._set_chan_name, chan_num),
-                           set_parser=str)
+                           set_parser=str,
+                           initial_value=f'Chan{chan_num:02d}')
 
-        self.add_parameter('synchronous',
-                           get_cmd=partial(self._parent._get_chan_synchronous, chan_num),
+        self.add_parameter('synchronous_enable',
```
FarBo, line 113: If you want the instrument is initilized with this parameter, rather than explicitly putting it in the __init__ as a class argument and the creating only readable parameter later, you can create a settable parameter and get it under __init__ like:
def __init__(...):
  .
  .
  .
  self.synchronous_enable()
And don't forget to change synchronous_enable to self.synchronous_enable() anywhere else you used it.

Then if you want it to be initilized with True, you just simply set it to its initial value:
 
```
self.add_parameter('synchronous_enable',
                           docstring='Is the channel in synchronous mode.',
                           get_cmd=None,
                           get_parser=bool,
                           set_cmd=partial(self._parent._set_chan_synchronous, chan_num),
                           initial_value=True
                           vals=vals.Bool())
```

what you have done seems duplication of creating an argument and excatly the same parameter, which is set in init and get as a parameter.

FarBo:
@edumur 

Thanks for the update. I only have one comment, If it is clarified, we can close this PR.

qcodes_contrib_drivers/drivers/Bilt/ITest.py
edumur: I wanted the user to have the possibility to change this parameter at the initialization of the device.
Is there another way of doing it?

FarBo: If you want to be forced at initialization, then there should not be any default value in it
Then you have it as an argument and pass it to the parameter under init, for example:

Def init(sync: bool):
  Self.enabled_synchronous(sync)


And you remove the initial value from the parameter

edumur: I guess it is done.
Please check if it is done in the way you suggested.

qcodes_contrib_drivers/drivers/Bilt/ITest.py
```diff
@@ -21,18 +21,11 @@ class iTestChannel(InstrumentChannel):
...
-                       chan_num: int,
-                       synchronous_enable: bool=True,
-                       synchronous_delay: float=1,
-                       synchronous_threshold: float=1e-5) -> None:
+                       chan_num: int) -> None:
         """
         Args:
             parent: The instrument to which the channel is attached.
             name: The name of the channel.
-            chan_num: The number of the channel in question
```
FarBo: Perhaps chan_num explanation should stay :)

edumur: oops

FarBo (approved):

PR #90, opened 2021-02-01
Update param names in Keysight E8267D similar to ones in qcodes driver Keysight E8267C

Akshita07:
Currently there is a driver for Keysight E8267C in Qcodes repository which has 4 parameters similar to Keysight E8267D driver but their names differ. This PR makes them consistent. 

- Updates param names in Keysight E8267D with same names as that of similar parameters in Keysight E8267C and deprecates old ones

@jenshnielsen @astafan8

### DIFF
--- a/qcodes_contrib_drivers/drivers/Keysight/Keysight_E8267D.py
+++ b/qcodes_contrib_drivers/drivers/Keysight/Keysight_E8267D.py
@@ -2,7 +2,7 @@ import numpy as np
 from qcodes import VisaInstrument, validators as vals
 from qcodes.utils.validators import Numbers
 from qcodes.utils.helpers import create_on_off_val_mapping
-
+from qcodes.instrument.parameter import DelegateParameter
 
 def parse_on_off(stat):
     if stat.startswith('0'):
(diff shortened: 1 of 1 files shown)


qcodes_contrib_drivers/drivers/Keysight/Keysight_E8267D.py
```diff
@@ -109,10 +115,16 @@ def __init__(self, name, address, step_attenuator=False, **kwargs):
...
+        self.add_parameter(name='output_rf',
+                           source=self.status,
+                           parameter_class=DelegateParameter)
         self.add_parameter(name='modulation_rf_enabled',
                            get_cmd='OUTP:MOD?',
                            set_cmd='OUTP:MOD {}',
                            val_mapping=on_off_mapping)
+        self.add_parameter(name='modulation_rf',
+                           source=self.modulation_rf_enabled,
+                           parameter_class=DelegateParameter)
```
astafan8, line 127: each of these parameters worth a comment in the code explaining "why"

qcodes_contrib_drivers/drivers/Keysight/Keysight_E8267D.py
```diff
@@ -142,10 +154,10 @@ def __init__(self, name, address, step_attenuator=False, **kwargs):
         self.connect_message()
 
     def on(self):
-        self.set('status', 'on')
+        self.set('output_rf', 'on')
 
     def off(self):
-        self.set('status', 'off')
+        self.set('output_rf', 'off')
```
astafan8, line 160: this should probably be improved further in order to get further away from string-specifying parameters:
```suggestion
        self.output_rf('off')
```

FarBo:
Can you post the link for qcodes driver? I don't see Keysight E8267C in qcodes drivers :(

qcodes_contrib_drivers/drivers/Keysight/Keysight_E8267D.py
jenshnielsen: If the aim is to make the parameters here consistent with the existing parameters in the other driver one should also consider doing this the other way around and leaving the existing parameters in this driver as deprecated delegates

Akshita07:
Here is the link to the driver in Qcodes .. https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument_drivers/agilent/E8267C.py

qcodes_contrib_drivers/drivers/Keysight/Keysight_E8267D.py
Akshita07: now, issuing deprecation warning which explains they reason why

astafan8 (approved):

PR #91, opened 2021-04-15
add driver for CopperMountain VNA M5180 with example

simonzihlmann:
This PR adds a driver and an example for the M5180 VNA from CopperMountain.

### DIFF
--- a/qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
+++ b/qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
@@ -3,9 +3,10 @@
 # Simon Zihlmannr <zihlmann.simon@gmail.com>, february/march 2021
 import logging
 import numpy as np
-from typing import Tuple
+from typing import Tuple, Optional
 
 from qcodes import VisaInstrument
+from qcodes.instrument import InstrumentBase
 from qcodes.utils.validators import Numbers, Enum, Ints
 from qcodes.utils.helpers import create_on_off_val_mapping
 
@@ -26,7 +27,7 @@ class FrequencySweepMagPhase(MultiParameter):
         start: float,
         stop: float,
         npts: int,
-        instrument: Optional[InstrumentBase] = "M5180",
+        instrument: InstrumentBase = "M5180,
         ) -> None:
         """
         Linear frequency sweep that returns magnitude and phase for a single
@@ -37,9 +38,7 @@ class FrequencySweepMagPhase(MultiParameter):
             start (float): Start frequency of linear sweep
             stop (float): Stop frequency of linear sweep
             npts (float): Number of points of linear sweep
-            instrument (InstrumentBase, optional): Instrument to which sweep
-            is bound to.
-                Defaults to "M5180".
+            instrument (InstrumentBase): Instrument to which sweep is bound to.
         """
         super().__init__(
             name,
@@ -113,7 +112,7 @@ class PointMagPhase(MultiParameter):
         start: float,
         stop: float,
         npts: int,
-        instrument: Optional[InstrumentBase] = "M5180",
+        instrument: InstrumentBase = "M5180",
         ) -> None:
         """Magnitude and phase measurement of a single point at start
         frequency.
@@ -506,21 +505,21 @@ class M5180(VisaInstrument):
         self.ask('*OPC?') # Wait for measurement to complete
 
         # Get data as string
-        freq = self.ask("SENS1:FREQ:DATA?")
-        s11 = self.ask("CALC1:TRAC1:DATA:FDAT?")
-        s12 = self.ask("CALC1:TRAC2:DATA:FDAT?")
-        s21 = self.ask("CALC1:TRAC3:DATA:FDAT?")
-        s22 = self.ask("CALC1:TRAC4:DATA:FDAT?")
+        freq_raw = self.ask("SENS1:FREQ:DATA?")
+        s11_raw = self.ask("CALC1:TRAC1:DATA:FDAT?")
+        s12_raw = self.ask("CALC1:TRAC2:DATA:FDAT?")
+        s21_raw = self.ask("CALC1:TRAC3:DATA:FDAT?")
+        s22_raw = self.ask("CALC1:TRAC4:DATA:FDAT?")
 
         # Get data as numpy array
-        freq = np.fromstring(freq, dtype=float, sep=',')
-        s11 = np.fromstring(s11, dtype=float, sep=',')
+        freq = np.fromstring(freq_raw, dtype=float, sep=',')
+        s11 = np.fromstring(s11_raw, dtype=float, sep=',')
         s11 = s11[0::2] + 1j*s11[1::2]
-        s12 = np.fromstring(s12, dtype=float, sep=',')
+        s12 = np.fromstring(s12_raw, dtype=float, sep=',')
         s12 = s12[0::2] + 1j*s12[1::2]
-        s21 = np.fromstring(s21, dtype=float, sep=',')
+        s21 = np.fromstring(s21_raw, dtype=float, sep=',')
         s21 = s21[0::2] + 1j*s21[1::2]
-        s22 = np.fromstring(s22, dtype=float, sep=',')
+        s22 = np.fromstring(s22_raw, dtype=float, sep=',')
         s22 = s22[0::2] + 1j*s22[1::2]
 
         return (freq, self._db(s11), np.angle(s11),
--- a/qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
+++ b/qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
@@ -6,7 +6,7 @@ import numpy as np
 from typing import Tuple, Optional
 
 from qcodes import VisaInstrument
-from qcodes.instrument import InstrumentBase
+from qcodes.instrument.base import InstrumentBase
 from qcodes.utils.validators import Numbers, Enum, Ints
 from qcodes.utils.helpers import create_on_off_val_mapping
 
@@ -27,7 +27,7 @@ class FrequencySweepMagPhase(MultiParameter):
         start: float,
         stop: float,
         npts: int,
-        instrument: InstrumentBase = "M5180,
+        instrument: "M5180",
         ) -> None:
         """
         Linear frequency sweep that returns magnitude and phase for a single
@@ -38,7 +38,7 @@ class FrequencySweepMagPhase(MultiParameter):
             start (float): Start frequency of linear sweep
             stop (float): Stop frequency of linear sweep
             npts (float): Number of points of linear sweep
-            instrument (InstrumentBase): Instrument to which sweep is bound to.
+            instrument: Instrument to which sweep is bound to.
         """
         super().__init__(
             name,
@@ -82,6 +82,7 @@ class FrequencySweepMagPhase(MultiParameter):
         Returns:
             Tuple[ParamRawDataType, ...]: magnitude, phase
         """
+        assert isinstance(self.instrument, M5180)
         self.instrument.write('CALC1:PAR:COUN 1') # 1 trace
         self.instrument.write('CALC1:PAR1:DEF {}'.format(self.name))
         self.instrument.write('CALC1:TRAC1:FORM SMITH')  # Trace format
@@ -90,10 +91,10 @@ class FrequencySweepMagPhase(MultiParameter):
         self.instrument.ask('*OPC?') # Wait for measurement to complete
 
         # get data from instrument
-        sxx = self.instrument.ask("CALC1:TRAC1:DATA:FDAT?")
+        sxx_raw = self.instrument.ask("CALC1:TRAC1:DATA:FDAT?")
 
         # Get data as numpy array
-        sxx = np.fromstring(sxx, dtype=float, sep=',')
+        sxx = np.fromstring(sxx_raw, dtype=float, sep=',')
         sxx = sxx[0::2] + 1j*sxx[1::2]
 
         return self.instrument._db(sxx), np.unwrap(np.angle(sxx))
@@ -112,7 +113,7 @@ class PointMagPhase(MultiParameter):
         start: float,
         stop: float,
         npts: int,
-        instrument: InstrumentBase = "M5180",
+        instrument: "M5180",
         ) -> None:
         """Magnitude and phase measurement of a single point at start
         frequency.
@@ -122,9 +123,7 @@ class PointMagPhase(MultiParameter):
             start (float): Start frequency (frequency at which it will measure)
             stop (float): Stop frequency, ideally start - 1 Hz
             npts (int): Number of points of the sweep, ideally 2
-            instrument (InstrumentBase, optional):  Instrument to which sweep
-            is bound to.
-            Defaults to "M5180".
+            instrument:  Instrument to which sweep is bound to.
         """
 
         super().__init__(
@@ -169,6 +168,7 @@ class PointMagPhase(MultiParameter):
         Returns:
             Tuple[ParamRawDataType, ...]: magnitude, phase
         """
+        assert isinstance(self.instrument, M5180)
         self.instrument.write('CALC1:PAR:COUN 1') # 1 trace
         self.instrument.write('CALC1:PAR1:DEF {}'.format(self.name[-3:]))
         self.instrument.write('CALC1:TRAC1:FORM SMITH')  # Trace format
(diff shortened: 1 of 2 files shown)


FarBo:
Hi @simonzihlmann 
Thank you for the contribution. We will try to review it as soon as possible.

In the meantime, you may try to solve the CI issues. It is mypy comlaining about types in the code, which you can find the issues if you click on the Details in any of above failed checks. Let us know if you need help finding and fixing those issue.

simonzihlmann:
Hi @FarBo
Thanks. I will have a look at the errors and try to solve them.

qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
```diff
@@ -0,0 +1,564 @@
...
+    """
+    Sweep that returns magnitude and phase.
+    """
+
+    def __init__(self,
+        name: str,
+        start: float,
+        stop: float,
+        npts: int,
+        instrument: Optional[InstrumentBase] = "M5180",
```
simonzihlmann, line 29: How do I have to declare the instrument class here? I guess not as a string. Is `instrument: Optional[InstrumentBase] = M5180,` fine?

Unfortunately, I have not found how to do it, neither in the example notebook nor in a similar driver in qcodes (ZNB driver from R&S).

qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
```diff
@@ -0,0 +1,564 @@
...
+        s11 = np.fromstring(s11, dtype=float, sep=',')
+        s11 = s11[0::2] + 1j*s11[1::2]
+        s12 = np.fromstring(s12, dtype=float, sep=',')
+        s12 = s12[0::2] + 1j*s12[1::2]
+        s21 = np.fromstring(s21, dtype=float, sep=',')
+        s21 = s21[0::2] + 1j*s21[1::2]
+        s22 = np.fromstring(s22, dtype=float, sep=',')
+        s22 = s22[0::2] + 1j*s22[1::2]
+
+        return (freq, self._db(s11), np.angle(s11),
```
simonzihlmann, line 526: Could you please explain me why I get the following error message (e.g. why are there different types)? `qcodes_contrib_drivers/drivers/CopperMountain/M5180.py:526:17: error: Incompatible return value type (got "Tuple[str, ndarray, Any, ndarray, Any, ndarray, Any, ndarray, Any]", expected "Tuple[ndarray]")
`

qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
```diff
@@ -0,0 +1,564 @@
...
+        freq = self.ask("SENS1:FREQ:DATA?")
+        s11 = self.ask("CALC1:TRAC1:DATA:FDAT?")
+        s12 = self.ask("CALC1:TRAC2:DATA:FDAT?")
+        s21 = self.ask("CALC1:TRAC3:DATA:FDAT?")
+        s22 = self.ask("CALC1:TRAC4:DATA:FDAT?")
+
+        # Get data as numpy array
+        freq = np.fromstring(freq, dtype=float, sep=',')
+        s11 = np.fromstring(s11, dtype=float, sep=',')
+        s11 = s11[0::2] + 1j*s11[1::2]
```
simonzihlmann, line 518: Why do I get the error `qcodes_contrib_drivers/drivers/CopperMountain/M5180.py:518:30: error: Unsupported operand types for * ("complex" and "str")`? Why is `s11` of type `str`?

simonzihlmann: This I figured out, I do not change the type of a variable any longer.

qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
simonzihlmann: Same as comment below.

qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
simonzihlmann: @FarBo could you maybe enlighten on how to solve this issue? Thanks in advance.

FarBo: Sure,

Do you really expect that another instrument other than M5180 uses this class? 
Also, this class uses instrument argument inside, so instrument cannot be Optional.

you probably can do such thing:

from qcodes.instrument import InstrumentBase
...
class PointMagPhase(MultiParameter):
...
instrument: InstrumentBase = "M5180"

simonzihlmann: Not really since it has some instrument specific commands. So I guess it would be better to hard code it. I'll try your suggestion. Thanks.

FarBo: Sorry, there was a little bit of mislead in my first suggestion

You can not use the instrument M5180 here because it needs to be an instance of  M5180 class, which means that there should be a name and address passed to it, and it is not possible in the driver level.

Actually, here you need to say:
**instrument: Optional['M5180'] = None** (please remember to import Optional from typing)
Then inside the init you should have:
**assert instrument is not None** 

The default value to instrument can be only None in this structure.

Since having this default is not going to serve what you want (as explained above), you can simply say:
**instrument: 'M5180'**

simonzihlmann: Hi @FarBo,
Thanks for your additional inputs. Though I am not sure if I fully understood your comment correctly. Here is what I have so far, but I still get some errors with mypy.
Code:
```
class FrequencySweepMagPhase(MultiParameter):
    """
    Sweep that returns magnitude and phase.
    """

    def __init__(self,
        name: str,
        start: float,
        stop: float,
        npts: int,
        instrument: "M5180",
        ) -> None:
        """
        Linear frequency sweep that returns magnitude and phase for a single
        trace.

        Args:
            name (str): Name of the linear frequency sweep
            start (float): Start frequency of linear sweep
            stop (float): Stop frequency of linear sweep
            npts (float): Number of points of linear sweep
            instrument: Instrument to which sweep is bound to.
        """
        super().__init__(
            name,
            instrument=instrument,
            names=(
                f"{instrument.short_name}_{name}_magnitude",
                f"{instrument.short_name}_{name}_phase"),
            labels=(
                f"{instrument.short_name} {name} magnitude",
                f"{instrument.short_name} {name} phase",
            ),
            units=("dB", "rad"),
            setpoint_units=(("Hz",), ("Hz",)),
            setpoint_labels=(
                (f"{instrument.short_name} frequency",),
                (f"{instrument.short_name} frequency",),
            ),
            setpoint_names=(
                (f"{instrument.short_name}_frequency",),
                (f"{instrument.short_name}_frequency",),
            ),
            shapes=((npts,), (npts,),),
        )
        self.set_sweep(start, stop, npts)

    def set_sweep(self, start: float, stop: float, npts: int) -> None:
        """Updates the setpoints and shapes based on start, stop and npts.

        Args:
            start (float): start frequency
            stop (float): stop frequency
            npts (int): number of points
        """
        f = tuple(np.linspace(int(start), int(stop), num=npts))
        self.setpoints = ((f,), (f,))
        self.shapes = ((npts,), (npts,))

    def get_raw(self) -> Tuple[ParamRawDataType, ParamRawDataType]:
        """Gets data from instrument

        Returns:
            Tuple[ParamRawDataType, ...]: magnitude, phase
        """
        self.instrument.write('CALC1:PAR:COUN 1') # 1 trace
        self.instrument.write('CALC1:PAR1:DEF {}'.format(self.name))
        self.instrument.write('CALC1:TRAC1:FORM SMITH')  # Trace format
        self.instrument.trigger_source('bus') # set the trigger to bus
        self.instrument.write('TRIG:SEQ:SING') # Trigger a single sweep
        self.instrument.ask('*OPC?') # Wait for measurement to complete

        # get data from instrument
        sxx = self.instrument.ask("CALC1:TRAC1:DATA:FDAT?")

        # Get data as numpy array
        sxx = np.fromstring(sxx, dtype=float, sep=',')
        sxx = sxx[0::2] + 1j*sxx[1::2]

        return self.instrument._db(sxx), np.unwrap(np.angle(sxx))
```

Error (they all appear in the get_raw method on the lines with `self.instrument`):
```
M5180.py:85: error: Item "None" of "Optional[InstrumentBase]" has no attribute "write"
M5180.py:86: error: Item "None" of "Optional[InstrumentBase]" has no attribute "write"
M5180.py:87: error: Item "None" of "Optional[InstrumentBase]" has no attribute "write"
M5180.py:88: error: Item "None" of "Optional[InstrumentBase]" has no attribute "trigger_source"
M5180.py:89: error: Item "None" of "Optional[InstrumentBase]" has no attribute "write"
M5180.py:90: error: Item "None" of "Optional[InstrumentBase]" has no attribute "ask"
M5180.py:93: error: Item "None" of "Optional[InstrumentBase]" has no attribute "ask"
M5180.py:99: error: Item "None" of "Optional[InstrumentBase]" has no attribute "_db"
```

Could you maybe help me to fix this?

jenshnielsen: Because of the way that that inheritance works mypy cannot see that in you case `self.instrument` is an instance of the M5180 but only that it (like in all MultiParameters) is either an instance of `InstrumentBase` or None. The easiest way to fix this is probably to help mypy a bit by asserting that this is really an `M5180`

I think that can be done by adding the following line to the start of your get_raw method

```
assert isinstance(self.instrument, M5180)
```

simonzihlmann: @jenshnielsen thanks a lot for the explanation!

FarBo:
@simonzihlmann 
Thanks you for the fix. I left a few minor comments otherwise it seems fine for me.
please resolve the conversations that are not pending anymore.
@jenshnielsen 
Could you have a final look and approve it if you find it acceptable?

simonzihlmann:
@FarBo Sorry, but I could not find your comments. Maybe I am blind..
I resolved the conversations.

qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
```diff
@@ -0,0 +1,572 @@
...
+        ) -> None:
+        """
+        Linear frequency sweep that returns magnitude and phase for a single
+        trace.
+
+        Args:
+            name (str): Name of the linear frequency sweep
+            start (float): Start frequency of linear sweep
+            stop (float): Stop frequency of linear sweep
+            npts (float): Number of points of linear sweep
```
jenshnielsen, line 40: npts is an int. Since the types are given in the signature you are also free to completely remove them from the docstring

FarBo:
@simonzihlmann 

Sorry, I forgot to press review button. Now you should be able to see the comments :)

jenshnielsen (approved):

qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
```diff
@@ -0,0 +1,572 @@
+# This Python file uses the following encoding: utf-8
+# Etienne Dumur <etienne.dumur@gmail.com>, august 2020
+# Simon Zihlmannr <zihlmann.simon@gmail.com>, february/march 2021
+import logging
+import numpy as np
+from typing import Tuple, Optional
+
+from qcodes import VisaInstrument
+from qcodes.instrument.base import InstrumentBase
```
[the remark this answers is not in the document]
simonzihlmann: removed

qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
simonzihlmann: true, changed to int

qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
```diff
@@ -0,0 +1,572 @@
...
+        self.add_parameter('averages',
+                           label='Averages',
+                           get_cmd='SENS1:AVER:COUN?',
+                           set_cmd='SENS1:AVER:COUN {}',
+                           get_parser=int,
+                           set_parser=int,
+                           unit='',
+                           vals=Numbers(min_value=1, max_value=999))
+
+        self.add_parameter(name='start',
```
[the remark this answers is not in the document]
simonzihlmann: Yes, it is a frequency. But since for a VNA people commonly use start, stop, span and center, I would like to leave it like this.
Furthermore, other vna drivers in qcodes also use this naming scheme, see here for example: https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument_drivers/rohde_schwarz/ZNB.py

@FarBo and @jenshnielsen: is there a naming scheme within qcodes for this kind of things? It would make sense to use the same name for the same parameters in all different drivers for the same kind of instruments (e.g. call the start frequency of all vnas start). This would allow to easily change between different physical instruments that do the same thing.

qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
```diff
@@ -0,0 +1,572 @@
...
+                           label='Start Frequency',
+                           get_parser=float,
+                           get_cmd='SENS1:FREQ:STAR?',
+                           set_cmd=self._set_start,
+                           unit='Hz',
+                           vals=Numbers(min_value=100e3,
+                                        max_value=18e9))
+
+        self.add_parameter(name='stop',
```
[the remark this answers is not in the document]
simonzihlmann: same as above

qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
```diff
@@ -0,0 +1,572 @@
...
+                           label='Stop Frequency',
+                           get_parser=float,
+                           get_cmd='SENS1:FREQ:STOP?',
+                           set_cmd=self._set_stop,
+                           unit='Hz',
+                           vals=Numbers(min_value=100e3,
+                                        max_value=18e9))
+
+        self.add_parameter(name='center',
```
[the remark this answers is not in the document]
simonzihlmann: same as above

qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
```diff
@@ -0,0 +1,572 @@
...
+                           label='Center Frequency',
+                           get_parser=float,
+                           get_cmd='SENS1:FREQ:CENT?',
+                           set_cmd=self._set_center,
+                           unit='Hz',
+                           vals=Numbers(min_value=100e3,
+                                        max_value=18e9))
+
+        self.add_parameter(name='span',
```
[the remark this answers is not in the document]
simonzihlmann: same as above

qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
```diff
@@ -0,0 +1,572 @@
...
+    def _set_start(self, val: float) -> None:
+        """Sets the start frequency and updates linear trace parameters.
+
+        Args:
+            val (float): start frequency to be set
+
+        Raises:
+            ValueError: If start > stop
+        """
+        stop = self.stop()
```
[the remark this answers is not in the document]
simonzihlmann: no, have not changed the name.

qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
```diff
@@ -0,0 +1,572 @@
...
+        if val >= stop:
+            raise ValueError("Stop frequency must be larger than start "
+                             "frequency.")
+        self.write("SENS1:FREQ:STAR {}".format(val))
+        # we get start as the vna may not be able to set it to the
+        # exact value provided.
+        start = self.start()
```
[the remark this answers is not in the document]
simonzihlmann: same as above

qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
```diff
@@ -0,0 +1,572 @@
...
+    def _set_stop(self, val: float) -> None:
+        """Sets the start frequency and updates linear trace parameters.
+
+        Args:
+            val (float): start frequency to be set
+
+        Raises:
+            ValueError: If stop < start
+        """
+        start = self.start()
```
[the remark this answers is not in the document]
simonzihlmann: same

qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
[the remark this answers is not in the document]
jenshnielsen: I don't think we have strong traditions. I agree that if it is obvious from the context (this being an vna)  (and unit) what this refers to it is fine to stick to short names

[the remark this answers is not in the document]
FarBo: The name of parameters should only reflect the behavior. Otherwise I think we do not have a naming scheme.
It also makes sense that similar instruments have similar parameter names for the same behavior. 
It is fine to leave the names as they are and I see that ZNB also using this naming scheme, which better to stick to that.

PR #98, opened 2021-06-21
ERASynth drivers

caenrigen:
Hi!

This is a driver using the new style implemented in https://github.com/QCoDeS/Qcodes/pull/3098.

- [x] I still need to add an example notebook.

For now I have a question, what is the proper way of making sure a certain command for a parameter was received by an instrument?

For this particular instrument I implemented the driver with manual control over the serial communication (instead of a Visa instrument) because from experience it was very unreliable in setting the parameters and some experiments require to as fast as possible and at the same time to be 100% sure that a certain parameter has been already set to the desired value.

To deal with this issue this driver always reads back the value it sets and sets it again until it succeeds... not very elegant nor ideal but at least makes it reliable.

Let me know your thoughts on this.

---

PS The link to the docs of this repo is a bit hidden on the top, and I think most people will be looking for it in the README. I  suggest in the PR to have it there as well.


FarBo:
@caenrigen 

Thank you for the driver. Regarding your question, I think you should use `ask_raw` method in the `Instrument` class. This method writes a command to the instrument and returns a response.
You should override this method since you are defining a new hardware communication.

I don't have a personal experience myself to guide you in detail, but I wanted to give you some insight. 
You could have a look at qcodes drivers, where some of them have used `ask_raw`, for instance this one:
https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument_drivers/Harvard/Decadac.py

Then you could probably use `ask_raw` to make your parameters set values reliable.

docs/conf.py
```diff
@@ -38,6 +38,7 @@
     'sphinx.ext.mathjax',
     'sphinx.ext.viewcode',
     'sphinx.ext.githubpages',
+    'qcodes.sphinx_extensions.add_parameter',
```
astafan8, line 41: let's make this dependency optional - contrib drivers does not restrict the qcodes version very much, hence it as to be flexible. so the guard could be ``try-except-ImportError`` or explicitly checkin for qcodes version > X.

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
```diff
@@ -0,0 +1,1014 @@
...
+See QCoDeS docs on writing drivers for more details.
+"""
+from __future__ import annotations
+
+from typing import Union
+import time
+import json
+from qcodes import Instrument
+from qcodes import validators
+from qcodes.instrument.base import InstanceAttr, add_parameter
```
astafan8, line 20: again, suggest to guard this with an import error, and inform the users that they need to have qcodes > X version installed.

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
```diff
@@ -0,0 +1,1014 @@
...
+import serial  # pip install pyserial
+from serial.tools.list_ports import comports  # pip install pyserial
```
astafan8, line 22: something like this:
```suggestion
try:
    import serial  # pip install pyserial
    from serial.tools.list_ports import comports  # pip install pyserial
except ImportError as e:
    raise ImportError("Failed to import ``pyserial`` package that is required for this driver, install it with ``pip install pyserial`` and try again") from e
```

FarBo:
@caenrigen 
We had a fix for doc building in the respoitory files, so I merged the master branch to have that fix for this PR. CI should not complain about that error anymore. But we have other problems that CI complains about. Let me know if I could help fixing those?

caenrigen:
> @caenrigen
> We had a fix for doc building in the respoitory files, so I merged the master branch to have that fix for this PR. CI should not complain about that error anymore. But we have other problems that CI complains about. Let me know if I could help fixing those?

Thank you @FarBo 

The rest of the errors depend on https://github.com/QCoDeS/Qcodes/pull/3098 being merged.

caenrigen:
@FarBo @astafan8 This one should also be ready for review now :)

caenrigen:
@FarBo @astafan8 I have updated this PR to make use of the latest qcodes and the new feature for documenting parameters.

Could you have a look and merge it?

If possible would appreciate if we can have a new release as well in order to be able to ship this new driver to PyPi

@AdriaanRol @dcrielaard

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
```diff
@@ -0,0 +1,902 @@
...
+    ) from e
+
+logger = logging.getLogger(__name__)
+
+BAUDRATE = 115200
+
+_CMD_TO_JSON_MAPPING: dict = {
+    # We will treat these differently when confirming the value
+    # because the instrument replies with a sentence containing the value.
+    # which is much faster than reading the full SJON
```
astafan8, line 40:
```suggestion
    # which is much faster than reading the full JSON
```

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
```diff
@@ -0,0 +1,902 @@
...
+
+    Example:
+
+    .. code-block::
+
+        from qcodes import Instrument
+        from qcodes_contrib_drivers.drivers.ERAInstruments import ERASynthPlus
+
+        Instrument.close_all()
+
```
astafan8, line 89: what is the reason for this? shall we leave this up to the users to close all instruments if they want/need it?
```suggestion
```

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
```diff
@@ -0,0 +1,902 @@
...
+            address: Used to connect to the instrument.
+                Run :meth:`.ERASynthBase.print_pyvisa_resources` to list available list.
+        """
+        super().__init__(name=name, address=address, terminator="\r\n", **kwargs)
+
+        # ##############################################################################
+        # Standard LO parameters
+        # ##############################################################################
+
+        # NB `initial_value` is not used because that would make the initialization slow
```
astafan8, line 144: note that initial_cache_value exists, if it can be useful for this parameter

astafan8 (approved):

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
caenrigen: Thanks for heads up! It is not used either because the instrument can be in an unknown state so better to have it as `None` in the beginning.

PS Overall the implementation of the serial communication of this instrument (in the physical instrument itself) is a bit messy and complicated to deal with. Unfortunately it was not developed with reliability and automation in mind 🥲

caenrigen:
@astafan8 thank you for the quick review, feel free to merge :)

FarBo:
@caenrigen 
Could you check the CI errors?
Seems mypy is not happe about bunch of lines in the code. Let me know if I could help passing it :)

caenrigen:
> @caenrigen
> Could you check the CI errors?
> Seems mypy is not happy about bunch of lines in the code. Let me know if I could help passing it :)

Hi! Just checked and tried to fix some of them :)

Please give me a hand if it does not pass again 😇

Kind regards,
Victor

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
```diff
@@ -0,0 +1,900 @@
...
+            "firmware": d_status["em"],
+        }
+
+    def clear_read_buffer(self) -> None:
+        """
+        Clears the read buffer. The instrument often sends information we do not care
+        about in this driver. This function discards the entire read buffer by reading
+        it.
+        """
+        while self.visa_handle.bytes_in_buffer:
```
FarBo, line 522: looking into visa_handle in qcodes, it refers to MessageBasedResource class in visa.resources.messagebased.
I can't find `bytes_in_buffer` attribute for this class, and I don't where this attribute is defined previously. 
Am I missing something?

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
```diff
@@ -0,0 +1,900 @@
...
+            name="debug_messages_en",
+            instrument=self,
+            vals=validators.Strings(),
+            initial_cache_value=None,
+            val_mapping={True: "1", False: "0"},
+            set_cmd="PD{}",
+        )
+        """Enables/disables debug printing on the serial port."""
+
+        self.visa_handle.baud_rate = BAUDRATE
```
FarBo, line 481: You are trying to add `baud_rate` attribute to `visa_handle` and set it to `BAUDRATE`. 
How can this be useful?
This is also one of mypy complains.

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
```diff
@@ -0,0 +1,900 @@
...
+
+        NB the read buffer is discarded before and after reading one line.
+        """
+        self.clear_read_buffer()
+        response = super().ask(f">{cmd}")
+        self.clear_read_buffer()
+
+        return response
+
+    def ask_raw(self, cmd: str) -> Union[Dict[str, str], str]:
```
FarBo, line 539: I am not sure how we can deal with this because mypy expects the type of the return to conform to the type of the `ask_raw` in the superclass, which is `str`.

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
caenrigen: It certainly worked while I was testing in a notebook

I guess it somehow routes to this:
https://pyvisa.readthedocs.io/en/latest/api/resources.html#pyvisa.resources.SerialInstrument.bytes_in_buffer

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
```diff
@@ -0,0 +1,900 @@
...
+        """
+        Turns OFF the RF output.
+        """
+        self.status(False)
+
+    # Custom communication
+
+    def get_idn(self) -> Dict[str, str]:
+        models: Dict[str, str] = {"0": "ERASynth", "1": "ERASynth+", "2": "ERASynth++"}
+        d_status: Dict[str, str] = self.get_diagnostic_status()
```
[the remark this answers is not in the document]
caenrigen: Not really, was trying to fix mypy as much as possible

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
caenrigen: I worked in this part a while ago, and do not fully recall this, I believe the idea was to set https://pyvisa.readthedocs.io/en/latest/api/resources.html?highlight=baud_rate#pyvisa.resources.SerialInstrument.baud_rate

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
```diff
@@ -0,0 +1,900 @@
...
+
+    def off(self) -> None:
+        """
+        Turns OFF the RF output.
+        """
+        self.status(False)
+
+    # Custom communication
+
+    def get_idn(self) -> Dict[str, str]:
```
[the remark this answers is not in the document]
FarBo: Please import Optional from typing

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
FarBo: OK. I wanted to make sure if everything is correct in the run time, then we can look into mypy thing

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
FarBo: OK. I wanted to make sure if everything is correct in the run time, then we can look into mypy thing

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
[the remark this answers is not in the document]
caenrigen: will do

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
caenrigen: Good point, it is a bit nasty, I will think a bit if I can implement it in some better way 🤔

caenrigen: I just double checked (executing the notebook with a debug print of `type(repsponse)`), the output of the function is always `str`, I fixed the type annotation, hope this solves the mypy issue

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
caenrigen: I double checked this as well:

Screenshot 2021-11-02 at 13 28 06

Screenshot 2021-11-02 at 13 27 30

caenrigen:
@FarBo I hope my last commit solves most issues

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
```diff
@@ -0,0 +1,900 @@
...
+
+    def get_configuration(self, par_name: str = None) -> Union[Dict[str, str], str]:
+        """
+        Returns the configuration JSON that contains all parameters.
+        """
+        config_json = json.loads(self._get_json("RA", "rfoutput"))
+
+        return config_json if par_name is None else config_json[par_name]
+
+    def get_diagnostic_status(self, par_name: str = None) -> Union[Dict[str, str], str]:
```
FarBo, line 619: Shouldn't the output be always a `str`, then?

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
FarBo: Then I think, `get_diagnostic_status` should be always `str` as commented below?

qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.py
caenrigen: no, it can be called manually with no arguments to get the JSON that the instrument returns for more "manual debugging"

caenrigen: PS if mypy is still not happy in the `ask_raw`, feel free to cast `str(self. get_diagnostic_status(...))`

FarBo:
@caenrigen 
I am working on mypy,

Will push a fix today :)

FarBo:
@caenrigen 
I pushed a few things to deal with mypy issues. seems the code is passing mypy check
there is a problem in the CI of the repositor, which is unrelated to this PR. That should be fixed, and this branch is updated with that fix to have the CI gets completed for this PR.
Could you test my latest commits to see if they didn't break anything?

caenrigen:
> @caenrigen
> I pushed a few things to deal with mypy issues. seems the code is passing mypy check
> there is a problem in the CI of the repository, which is unrelated to this PR. That should be fixed, and this branch is updated with that fix to have the CI gets completed for this PR. I will handle that.
> Could you test my latest commits to see if they didn't break anything?

Thank you for the help, fixed it with my last commit (the attribute needs to be queried every time)

FarBo (approved):

caenrigen:
Thank you for the review and help!

Quick question, when would it be possible to have a release?

FarBo:
No worries

There is abit of work left for the release step in the CI. Once that is done, we will release a version right after.

caenrigen:
> No worries
> 
> There is abit of work left for the release step in the CI. Once that is done, we will release a version right after.

Awesome! Thanks :)

PR #99, opened 2021-06-24
Add marker configuration capabilities

alexsludds:
For my work with M3202A modules in a 9010 chassis I need to use markers for precise timing of other equipment in the setup. This function allows me to generate marker signals from the trigger port.


qcodes_contrib_drivers/drivers/Keysight/SD_common/SD_AWG.py
```diff
@@ -283,6 +283,35 @@ def set_trigger_io(self, value: int, verbose: bool = False) -> Any:
...
+            nAWG=channel_number,
+            markerMode=markerMode,
+            trgPXImask=trgPXImask,
+            trgIOmask=trgIOmask,
+            value=markerValue,
+            syncMode=syncMode,
+            length=length,
+            delay=delay
+        )
+        return result_parser(result, verbose)
```
FarBo, line 314: I think you forgot to correctly use `result_parser ` function. It requires 3 args in the function construct, which the second arg. is an string intended to pass the name of what you want to parse:
`return result_parser(result, "pass here whatever name you like", verbose)`

alexsludds: Thanks for this message! I misunderstood what result parser was doing. I submitted a new piece of code which has removed it. This code is working on my hardware in the lab.

FarBo: Sure.
So, verbose is not applicabale anymore with this change.

FarBo:
@alexsludds 
We had a fix for doc building in the respoitory files, so I merged the master branch to have that fix for this PR. Now the CI passes

alexsludds:
Awesome! Can we merge this now?

FarBo:
@alexsludds 
We have still an open conversation. in your function, you have 'verbose' argument that you are not using it after you recent change. This should be corrected before any merge :)

qcodes_contrib_drivers/drivers/Keysight/SD_common/SD_AWG.py
astafan8: removed verbose argument

astafan8 (approved):

PR #102, opened 2021-08-17
Update M5180.py

erikawa-e:
"assert isinstance(self.instrument, M5180)"
I received some error from this line.
Is it because I am using S5045 instead of M5180?

There must be a better way to code so that this driver can accommodate all the CopperMountain VNAs.
I am not good enough to do it but if some one can tell me how to do, I will try to work on it.

### DIFF
--- a/qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
+++ b/qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
@@ -167,7 +167,7 @@ class PointMagPhase(MultiParameter):
         Returns:
             Tuple[ParamRawDataType, ...]: magnitude, phase
         """
-        assert isinstance(self.instrument, M5180)
+
         self.instrument.write('CALC1:PAR:COUN 1') # 1 trace
         self.instrument.write('CALC1:PAR1:DEF {}'.format(self.name[-3:]))
         self.instrument.write('CALC1:TRAC1:FORM SMITH')  # Trace format


FarBo:
@erikawa-e
This driver is developed and tested for the model M5180. Although it might be used for other models, but it should be done after comparing another model's manual to see if the driver is compatible for that model.

As far as I looked, the driver doesn't bound the driver to the actual model of the VNA. Meaning the instrument calss name of M5180 is just a name that can be utilized for this model or another model (if this driver be compatible with that).

So I have a few questions:
1. How do you initialize the driver?
2. When does the error pops up and what is the complete error message?

erikawa-e:
Thank you so much for writing this driver first of all!!
It saved my day(s)!!

After removing "assert isinstance(self.instrument, M5180)", the driver is working fine for S5045 (as far as I could observe).
(Of course, I had to change the frequency and power ranges.)

>As far as I looked, the driver doesn't bound the driver to the actual model of the VNA. Meaning the instrument calss name of M5180 is just a name that can be utilized for this model or another model (if this driver be compatible with that).

This is what I also thought.
I might be doing something strange somewhere..

>How do you initialize the driver?
from qcodes.instrument_drivers.CopperMountain.M5180 import M5180
vna = M5180(name='vna', address='TCPIP0::localhost::hislip0::INSTR')

>When does the error pops up and 
vna.S21()

>what is the complete error message?
AssertionError: failed (as far as I remember)
(Sorry that now a long measurement is running and so I will come back to you with complete error messages later.)

erikawa-e:
After restarting Qcodes, the error message disappeared.
I am sorry that it might be doing something wrong with my initialization and so I cancel this request.
Thank you again for your wonderful driver and I think this driver can be used for any models from CopperMountain which works with "connection software S2VN".

FarBo:
@erikawa-e 

That's good and expected news because as I mentioned the driver does not bound the code to the actual model of VNA.
By the way, the driver is developed by @simonzihlmann and we only reviewed and helped to get it merged :)

PR #114, opened 2022-01-06
Add SignalCore driver

edumur:
All in the title I guess ^^ .

### DIFF
--- a/docs/examples/SignalCore.ipynb
+++ b/docs/examples/SignalCore.ipynb
@@ -1,0 +1,207 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+  "nbsphinx": {
+   "execute": "never"
+  },
+  },
+   "source": [
+    "# Example with the microwave source SC5521A from Signal Core"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Logging hadn't been started.\n",
+      "Activating auto-logging. Current session state plus future input saved.\n",
+      "Filename       : C:\\Users\\ed261978\\.qcodes\\logs\\command_history.log\n",
+      "Mode           : append\n",
+      "Output logging : True\n",
+      "Raw input log  : False\n",
+      "Timestamping   : True\n",
+      "State          : active\n",
+      "Qcodes Logfile : C:\\Users\\ed261978\\.qcodes\\logs\\210604-4564-qcodes.log\n"
+     ]
+    }
+   ],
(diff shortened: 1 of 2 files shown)


edumur:
That's interesting, mypy passed on my laptop:
```bash
python -m mypy --ignore-missing-import --no-warn-unused-ignores -p qcodes_contrib_drivers
Success: no issues found in 156 source files
```

astafan8:
> That's interesting, mypy passed on my laptop:

likely because you have Windows? and the CI is using linux. mypy understands platform differences so we can fix it, let me find out how ....

qcodes_contrib_drivers/drivers/SignalCore/SignalCore.py
```diff
@@ -0,0 +1,311 @@
...
+        name (str): Name of the instrument.
+        dll_path (str): Path towards the instrument DLL.
+        """
+        
+        (super().__init__)(name, **kwargs)
+        
+        # Adapt the path to the computer language
+        dll_path = os.path.join(os.environ['PROGRAMFILES'], dll_path)
+
+        self._dll = ctypes.WinDLL(dll_path)
```
astafan8, line 134: so let's use this https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks here - and if the platform is not windows then raise a nice error to the user (mypy will also like it and understand it)

docs/examples/SignalCore.ipynb
```diff
@@ -0,0 +1,207 @@
+{
```
astafan8, line 1: there seems to be an issue with the notebook as well that it's not a valid JSON, see here https://github.com/QCoDeS/Qcodes_contrib_drivers/blob/d8b66cd63692e22267ec29a2910516f8ce5cb18d/docs/examples/SignalCore.ipynb

edumur: Yep that's me adding:
```json

            "nbsphinx":{
               "execute":"never"
            }
```
too fast...
Thanks for the catch!

qcodes_contrib_drivers/drivers/SignalCore/SignalCore.py
edumur: Nice, I am implementing it, thanks!

edumur:
Well, I tried my best (can't make local test first)

edumur:
ok, I am out of ideas...
Does someone has a hint?

qcodes_contrib_drivers/drivers/SignalCore/SignalCore.py
astafan8: should be done now

astafan8 (approved):

edumur:
Well, great thank for all that work!

PR #116, opened 2022-02-10
remove redundant typecast

jenshnielsen:
With newer versions of qcodes the typing of ChannelList has improved to reflect that `__getitem__` always returns a InstrumentChannel when called with an int so this cast is now redundant


jenshnielsen:
@jpsecher For the QDAC2 this method was typed as returning a sequence but does actually return a numpy array

Unfortunately numpy arrays are currently not considered sequences see https://github.com/numpy/numpy/issues/2776 

Do you have a preference if the type signature should change like I have done here or we should convert the output to a list

jenshnielsen:
@sldesnoo-Delft @eendebakpt  

Numpy 1.22 triggers a type checking error in the m4i driver related to from buffer. Since I don't have much context on this driver I have silenced the warning here but perhaps you want to look into a better fix?

sldesnoo-Delft:
I will have a look at the M4i driver. This week I already worked on another fix for the driver.

sldesnoo-Delft:
I just learned that numpy 1.22 requires Python 3.8. However Python 3.8 is not supported by the older Keysight PXI system drivers and firmware. It requires the latest drivers and firmware that have been released in December 2021.
I've not yet had time to test the new firmware and update the drivers.
This might imply a breaking change for people with older licenses that do not allow an update of the firmware.
Dropping support for Python 3.7 would mean that people have to buy new licenses to be able to upgrade to new firmware and drivers.

What's the policy of qcodes regarding the supported Python versions?

astafan8 (approved):

jpsecher:
@jenshnielsen thanks, the changes to QDAC2.py looks fine.

jenshnielsen:
@sldesnoo-Delft this does not actually drop 3.7 support just yet but just stops type checking with 3.7. We were going to drop 3.7 exactly because numpy, pandas, xarray ipython and so on has dropped support. Could you copy your comment into https://github.com/QCoDeS/Qcodes/pull/3889 that may be an important datapoint for delaying the dropping of 3.7 but its nice if we can collect all info there.

Issue #120, opened 2022-03-02
CopperMountain M5180 -- typo in __init__ docstring?


omajoshi:
https://github.com/QCoDeS/Qcodes_contrib_drivers/blob/2fd9c391458417e3914b0ca4495ffb4174497bf1/qcodes_contrib_drivers/drivers/CopperMountain/M5180.py#L207

Every line in the file says M5180 except this one, which says S5180. Is that just a typo in the `__init__` docstring?

jenshnielsen:
I am not very familar with this driver but that does seem likely. Are you willing to open a pull request to fix that?

omajoshi:
Going to test the driver on the M5180 first and will make the fix if it works.

omajoshi:
I'm going to compare the driver to the documentation for the M5180, and I'll will make the fix if everything checks out.

simonzihlmann:
I confirm that it is a typo in the docstring and that the driver has been written for the M5180 model. I can fix it at the same time when I propose some changes (I have some minor changes (updated validators, changes to the PointMagPhase class) that I have not yet put into a PR.  Try to do this soonish.

jenshnielsen:
Thanks @simonzihlmann For some reason the original commits no longer map to your github user name (did you change your email perhaps) So I was struggeling with remembering how to mention you but great that you saw it

simonzihlmann:
> Thanks @simonzihlmann For some reason the original commits no longer map to your github user name (did you change your email perhaps) So I was struggeling with remembering how to mention you but great that you saw it

Not on purpose. But it could be that I used a different email address when committing than the one that github knows about. I have added the second one to my profile. Not sure if it has something to do with this though.

jenshnielsen:
Hmn not sure what happened anyway its showing correctly now https://github.com/QCoDeS/Qcodes_contrib_drivers/commits/master/qcodes_contrib_drivers/drivers/CopperMountain/M5180.py?author=simonzihlmann

PR #121, opened 2022-03-04
Feat cm m5180 improvements

simonzihlmann:
Improvements for the PointMagPhase class that mimics a CW mode. Now the averages over all points in a trace (generally two as this is the minimum) is used instead of throwing away one data point (half the data).
Validators have been updated.
This also fixes: https://github.com/QCoDeS/Qcodes_contrib_drivers/issues/120

### DIFF
--- a/qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
+++ b/qcodes_contrib_drivers/drivers/CopperMountain/M5180.py
@@ -3,14 +3,16 @@
 # Simon Zihlmannr <zihlmann.simon@gmail.com>, february/march 2021
 import logging
 import numpy as np
-from typing import Tuple, Optional, Any
+import cmath, math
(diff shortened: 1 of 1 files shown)


omajoshi:
You might want to look into https://coppermountaintech.com/help-s2/calcform.html?q=SLOG for different data formats to reduce the number of conversions from rectangular to polar, etc.

simonzihlmann:
> You might want to look into https://coppermountaintech.com/help-s2/calcform.html?q=SLOG for different data formats to reduce the number of conversions from rectangular to polar, etc.

Thanks for the hint. I might consider in a future improvement of the driver. For the moment it works fine.

simonzihlmann:
@jenshnielsen Am I missing anything?

jenshnielsen:
Sorry this slipped throught the cracks

PR #123, opened 2022-03-10
Move CM S5048 driver to the correct folder

omajoshi:
Rename qcodes_contrib_drivers/drivers/CMTS5048.py to qcodes_contrib_drivers/drivers/CopperMountain/S5048.py


astafan8 (approved):
shall the old .py file be left in its original location with a deprecation warning and importing all the public items (e.g. driver class) from the new .py file?

PR #127, opened 2022-04-22
Two 2D scan examples & refuse non-identifier instrument name

jpsecher:


qcodes_contrib_drivers/drivers/QDevil/QDAC2.py
```diff
@@ -2170,3 +2171,10 @@ def _set_up_manual_triggers(self) -> None:
     def _set_up_simple_functions(self) -> None:
         self.add_function('reset', call_cmd='*rst')
         self.add_function('abort', call_cmd='abor')
+
+    def _check_instrument_name(self, name: str) -> None:
```
astafan8, line 2175: stricktly qcodes does not prescribe this requirement for instrument names. but perhaps it should? could you explain when did it become a problem for qdac?

astafan8 (approved):

astafan8 (commented):
for the new notebooks, don't forget to disable execution in the docs build, read here how to do that https://qcodes.github.io/Qcodes/examples/writing_drivers/Creating-Instrument-Drivers.html#Documentation

qcodes_contrib_drivers/drivers/QDevil/QDAC2.py
jpsecher: The `doNd` functions break when the instrument name is not `isidentifier()`

qcodes_contrib_drivers/drivers/QDevil/QDAC2.py
```diff
@@ -1815,6 +1815,7 @@ def __init__(self, name: str, address: str, **kwargs) -> None:
             **kwargs: additional argument to the Visa driver
         """
         super().__init__(name, address, terminator='\n', **kwargs)
+        self._check_instrument_name(name)
```
jenshnielsen, line 1818: It would probably be better to do this before the call to `__init__` otherwise you will leave around a partially setup instrument.

PR #129, opened 2022-05-19
Add Holzworth HS9008B, General Microwave GM349, and NI DAQ drivers

ty-zhao:
Slightly modified Holzworth HS900 driver so that it now works properly for Holzworth HS9008B signal generators. Modified National Instrument DAQ driver so that digital output devices can now be properly controlled. Tested with NI PXI-6509.

### DIFF
--- a/qcodes_contrib_drivers/drivers/GeneralMicrowave/GM349.py
+++ b/qcodes_contrib_drivers/drivers/GeneralMicrowave/GM349.py
@@ -1,0 +1,105 @@
+# -*- coding: UTF-8 -*-
+# Tongyu Zhao <ty.zhao.work@gmail.com> summer 2022
+"""
+Driver for Kratos General Microwave Series 349 and 349H Attenuators.
+
+These attenuators are passive devices that adjust attenuation
+based on digital signals supplied to their J3 connector.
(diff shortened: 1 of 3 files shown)


astafan8 (approved):

simonzihlmann:
@ty-zhao May I ask what you had to change to make the driver working with the model HS9008B?

ty-zhao:
> @ty-zhao May I ask what you had to change to make the driver working with the model HS9008B?

Just a minor change. On line 246 of the original HS900 driver, I changed `terminator='\n'` to `termintor=''`. The return strings of our HS9008B don't have '\n' as the line terminator. Using the original driver would cause a timeout error.

simonzihlmann:
> Just a minor change. On line 246 of the original HS900 driver, I changed `terminator='\n'` to `termintor=''`. The return strings of our HS9008B don't have '\n' as the line terminator. Using the original driver would cause a timeout error.

I see. But for that you do not need to change the driver. You could just pass the terminator string (in your case "") in the init-function when you initialize your device, no?

simonzihlmann:
I see. Well then I guess it is complicated to have a single file only.

PR #130, opened 2022-05-24
Analog modulation for NI RFSG driver

mgunyho:
Add functionality to enable analog (amplitude / frequency / phase) modulation for NI-RFSG based devices. Includes example code in the example notebook. Tested amplitude modulation with a PXIe-5654.


astafan8 (approved):
thank you!

PR #131, opened 2022-05-24
AFG3000: Use on_off_val_mapping instead of custom vals and get_parser

mgunyho:
Before this change, setting boolean parameters to `True`/`False` failed silently:

```python
from qcodes_contrib_drivers.drivers.Tektronix.AFG3000 import AFG3000

afg = AFG3000("AFG", address=...)

afg.state_output1.set(1)
assert afg.state_output1() is True  # ok

afg.state_output1.set(False)  # no error
assert afg.state_output1() is False  # fails, afg.state_output1 is still True !

# this works
afg.state_output1.set(0)
assert afg.state_output1() is False
```

With this patch, the second assert above also succeeds.

Note: while this is a bug in this driver, interestingly `Enum(0, 1).validate(True)` succeeds (works also with floats 0.0 and 1.0). Seems like this is due to `hash(1) == hash(1.0) == hash(True) == 1`. Perhaps this is a bit surprising, I would expect `Enum` to be a bit more strict. But maybe it's enough to warn about this in the documentation.

### DIFF
--- a/qcodes_contrib_drivers/drivers/Tektronix/AFG3000.py
+++ b/qcodes_contrib_drivers/drivers/Tektronix/AFG3000.py
@@ -2,6 +2,8 @@ from typing import Any, Tuple
 
 from qcodes import VisaInstrument
 import qcodes.utils.validators as vals
+from qcodes.utils.helpers import create_on_off_val_mapping
+
 
 class AFG3000(VisaInstrument):
     """Qcodes driver for Tektronix AFG3000 series arbitrary function generator.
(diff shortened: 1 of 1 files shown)


astafan8:
@mgunyho 

> Note: while this is a bug in this driver, interestingly Enum(0, 1).validate(True) succeeds (works also with floats 0.0 and 1.0). Seems like this is due to hash(1) == hash(1.0) == hash(True) == 1. Perhaps this is a bit surprising, I would expect Enum to be a bit more strict. But maybe it's enough to warn about this in the documentation.

indeed :) i haven;t looked at it, but it might be related to the fact that in python ``True == 1`` etc, as you also mention. so we can just keep it that.

astafan8 (approved):

PR #137, opened 2022-06-13
QDevil QDAC2: BUS and HOLD take up internal triggers 15 and 16

jpsecher:


astafan8:
@jpsecher with the latest qcodes release (which is used by CI) the name of the instruments has to be a valid python indetifier, and that's not the case for qdac tests - could you have a look and fix that? could be in this same PR or a separate one

jpsecher:
Hi, can I get you to accept these changes which includes the fix that makes the driver work with the new version of QCoDeS?

astafan8 (approved):

PR #141, opened 2022-07-18
Add SC5511B driver

ttshaw1:
Added driver for SignalCore SC5511B. Plenty of room for optimizations and new features but might be enough to get some other group up and running.


astafan8 (approved):
tank you @ttshaw1 for the contribution!

i only changed the driver to use the instrument instance's own logger (here's a bit on how to use it https://qcodes.github.io/Qcodes/examples/logging/logging_example.html#Filtering-log-messages-by-instrument ).

it would be great to make private all the methods that are actually gets/sets of the parameters, but it's ok :)

i'll make a new release of qcodes_contrib_drivers after this PR merges so that you can install this driver as part of qcodes_contrib_drivers using `pip install qcodes_contrib_drivers`

astafan8:
@ttshaw1 could you have a look at these typing errors https://github.com/QCoDeS/Qcodes_contrib_drivers/runs/7405036031?check_suite_focus=true ? if haven't worked with mypy before, we can do one of the following: exclude this driver from type checking alltogether, add ``# type: ignore`` to all the troubling lines, remove existing partial type annotations. I don't recommend either of these, but we have to be realistic and do what we can.

jenshnielsen:
@ttshaw1 is there still interest in this driver? It would need a bit of work to get the CI to pass

PR #143, opened 2022-08-16
Mypy better missing imports

jenshnielsen:
This means that mypy will only skip missing imports for modules that we have explicitly marked as known to be missing or non typed. This also found a bug in a test file. These tests dont run automatically since they require the instrument being present


jenshnielsen:
Note that this will not pass until https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/142 is merged

WilliamHPNielsen (approved):

PR #154, opened 2022-09-07
QDevil QDAC2: Update to agree with new firmware

jpsecher:
- Accept shorter version numbers
- Sleep after reset (*RST)
- Update examples


requirements.txt
```diff
@@ -1 +1,2 @@
 mypy==0.971
```
jenshnielsen, line 1: This file is only used to constant the requirements to specific versions. Please at to setup.cfg (under test)

qcodes_contrib_drivers/tests/QDevil/readme.md
```diff
@@ -27,3 +27,4 @@ Static types:
     $ source venv/bin/activate
     $ pip install --upgrade pip
     $ pip install qcodes pytest pyvisa-sim pyvisa-py mypy
+    $ pip install -e .
```
jenshnielsen, line 30: The aim of how our tests are set up hopefully it should be enough to do `pip install -e .[test]` if any deps are missing we should add them there

jpsecher: Ahh, I did not know that you could write `pip install -e .[test]`

astafan8 (approved):

PR #158, opened 2022-09-20
QDevil QDAC2: Leakage sweep

jpsecher:
- set_virtual_voltage affects whole arrangement


docs/examples/QDevil/QDAC2/GateLeakage.ipynb
```diff
@@ -0,0 +1,136 @@
+{
```
astafan8, line 1: looks like the execution of this example during the docs build should be disabled, see here https://qcodes.github.io/Qcodes/examples/writing_drivers/Creating-Instrument-Drivers.html#Documentation

qcodes_contrib_drivers/tests/QDevil/test_sim_qdac2_leakage.py
```diff
@@ -0,0 +1,129 @@
...
+        'sens:nplc 2,(@1,2,3)',
+        'read? (@1,2,3)',
+        'read? (@1,2,3)',
+        'sour1:volt:mode fix',
+        'sour1:volt 0.3',
+        'sour2:volt:mode fix',
+        'sour2:volt 0.0',
+        'sour3:volt:mode fix',
+        'sour3:volt 0.4',
+    ]
```
astafan8, line 126: some of the tests fail due to being sensitive to floating point precision, perhaps for the lines with numbers some code can be written to take the number and compare it to expected using ``pytest.approx`` or some other method?

astafan8 (approved):

jpsecher:
After talking with Fabrizio, Torbjørn, Joost, etc at NBI, we decided to call output channels for "contacts" to avoid confusion of what an "arrangement" can contain.

PR #177, opened 2022-11-16
Use cache from python github actions

jenshnielsen:
* Reflects change in qcodes. 
* Less code to maintain
* Avoids deprecation warnings


jenshnielsen:
Should be merged after https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/178 so cache looks at the right file

astafan8 (approved):

PR #179, opened 2022-11-21
Fix Keithley6500 driver

eendebakpt:
The `Keithley_Sense` submodule defined the `label` and `unit` of the `measure` parameter incorrect. We have to set them as a string, not as a method.


jenshnielsen (approved):

jenshnielsen:
thanks

PR #180, opened 2022-11-24
Sync multiple QDAC-IIs together

jpsecher:


jpsecher:
So I guess I am ready for release.

The driver for multiple QDACs does not yet behave as a proper Visa instrument, and I will work on an update later when I figure out the most compatible way to achieve that.  But I think it is important that we have release now so that people can start using the synchronisation feature.

PR #182, opened 2022-12-09
Fix marker period start mnemonic

maij:
While the QDAC II manual states "PST" as the abbreviated mnemonic, when using this command the QDAC raises an error with unknown mnemonic. Using the unabbreviated form works.

I haven't seen this issue for other commands, but have not checked thoroughly.

N.B. I am using firmware version 7-0.17.5

### DIFF
--- a/qcodes_contrib_drivers/drivers/QDevil/QDAC2.py
+++ b/qcodes_contrib_drivers/drivers/QDevil/QDAC2.py
@@ -526,7 +526,7 @@ class _Waveform_Context(_Channel_Context):
     def _period_start_marker(self, wave_kind: str) -> QDac2Trigger_Context:
         if not self._marker_period_start:
             self._marker_period_start = self.allocate_trigger()
-        self._write_channel(f'sour{"{0}"}:{wave_kind}:mark:pst {self._marker_period_start.value}')
+        self._write_channel(f'sour{"{0}"}:{wave_kind}:mark:pstart {self._marker_period_start.value}')
         return self._marker_period_start
 
     def _make_ready_to_start(self, wave_kind: str) -> None:


astafan8:
@jpsecher perhaps this should be reviewed by you? :)

jpsecher:
Yes, unfortunately there is a discrepancy where some of the short form commands (in firmware up to version 0.17.5) do not follow the SCPI standard.  This is fixed in more recent firmware, but this newer firmware has not been released yet.  I will make sure that it will be released as soon as possible and post a link here.

In the meantime, you can use the long form of the keyword (like `pstart`) or use the short form with an extra vowel (like `psta`).  Sorry for the inconvenience.

jpsecher:
The newest firmware is here:

    https://qdevil-public.s3.eu-west-1.amazonaws.com/qdac2/qdac2-fw-update_11-1.14.exe
    https://qdevil-public.s3.eu-west-1.amazonaws.com/qdac2/qdac2-fw-update-macos_11-1.14
    https://qdevil-public.s3.eu-west-1.amazonaws.com/qdac2/qdac2-fw-update-linux_11-1.14

but, out of the box, it will not be usable with QCoDeS QDac2 drivers prior to version 1.1.2 (which is not released yet), see https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/180

jenshnielsen:
@jpsecher Just to be sure the long form like in this pr will continue to be supported so its safe to merge this?

jenshnielsen:
@maij It looks like some of the tests are written to assume the short from and would need updating

jpsecher:
The test have been updated together with the fix, so I don't know why the old tests are run.  Can I retrigger the tests?

jenshnielsen:
@jpsecher I guess the tests was updated on #180 ? If that one is ready for review it perhaps makes more sense to merge that and do a new release and close this one ?

jpsecher:
But I need to lower the accepted version in the driver then (because it 
now actually works with all versions).  I will do that right away.

/JP

maij:
Thanks for following up -- happy to close this once the other changes are merged.

PR #183, opened 2022-12-12
Pin packaging to < 22 for drivers using legacy versions and pin ipython to 8.6

jenshnielsen:


jenshnielsen:
@jpsecher The QDAC II version comparison logic makes use of legacy versions which have been removed in packaging 22.0. Here I have pinned it to an older version but it would be nice to update the logic to not have this issue.

In QCoDeS we implemented some logic that modifies version numbers to be consistent. Alternatively you may have to implement your own logic.

astafan8 (approved):

jpsecher:
I have implemented my own version logic (because the first part of the 
qdac2 version concerns the fpga and has no effect on the driver).

PS: I just realised that I have been replying to these messages from 
Github by hitting Reply, which then wrongly goes to Jens Hedegaard 
Nielsen ***@***.***>, so I guess you never got any of the 
emails I sent, I guess...

Cheers,
/JP

jenshnielsen:
Cool we can remove the pin once #180 is removed. 
I was under the impression that if you reply to a email comment they should normally end up here too (but perhaps you need reply all) I don't use email for my github notefications personally so a bit unsure

jpsecher:
FYI it was a problem with Thunderbird that for some reason has changed 
the default setting.

Cheers,
/JP

PR #191, opened 2022-12-28
add driver for HP8594E

RasmusBC59:
adds QCoDeS driver for the Hewlett Packard HP8594E Network Analyzer


qcodes_contrib_drivers/tests/HP/test_HP8594E.py
```diff
@@ -0,0 +1,31 @@
...
+import numpy as np
+from qcodes_contrib_drivers.drivers.HP.HP8594E import HP8594E
+
+
+@pytest.fixture(scope="function")
+def driver():
+    HP_sim = HP8594E(
+        "HP_sim",
+        "GPIB::1::INSTR",
+        pyvisa_sim_file="qcodes_contrib_drivers.sims:HP8594E.yaml",
```
jenshnielsen, line 11: This feature requires at least qcodes 0.36, could you bump the min version in pyproject.toml to match that?
We might as well make it 0.36.1 since that contains an important bug fix

qcodes_contrib_drivers/drivers/HP/HP8594E.py
```diff
@@ -0,0 +1,207 @@
...
+from typing import Any, Dict
+import struct
+import numpy as np
+import datetime
+import qcodes.utils.validators as vals
+from qcodes.instrument.parameter import (
+    Parameter,
+    ParameterWithSetpoints,
+    ParamRawDataType,
+)
```
jenshnielsen, line 11: I have been doing some refactoring of the qcodes public interface. Could you update this to import paramters from qcodes.parameters and the VisaInstrument from qcodes.instrument. 

This requires qcodes 0.35 but since I already proposed bumping this below it should be fine

qcodes_contrib_drivers/drivers/HP/HP8594E.py
```diff
@@ -0,0 +1,207 @@
...
+from qcodes.instrument.visa import VisaInstrument
+import numpy.typing as npt
+
+log = logging.getLogger(__name__)
```
jenshnielsen, line 15: I can't see that this logger is used anywhere?

qcodes_contrib_drivers/tests/HP/test_HP8594E.py
RasmusBC59: Hi sorry for the late response, but have been on vacation the last three weeks.
Change implemented.

qcodes_contrib_drivers/drivers/HP/HP8594E.py
RasmusBC59: Change implemented.

qcodes_contrib_drivers/drivers/HP/HP8594E.py
RasmusBC59: removed unused logger

jenshnielsen (approved):

PR #203, opened 2023-02-09
upgrade mypy to 1.0.0

jenshnielsen:
and use ~= for requirements

I do not know why dependabot is not triggering this


jenshnielsen:
This also found a use of an undefined variable

astafan8 (approved):

PR #209, opened 2023-03-20
QDAC1: Allow slopes greater than 10V

jpsecher:
- Handle excess output
    - Report errors in complex commands


astafan8 (approved):

jpsecher:
Hmm, I just tried the same trick on the QDAC2 driver, and it did not work as expected, that is, the timeout was not overridden...

jpsecher:
Of course it is in seconds (https://www.youtube.com/watch?v=QfnC_3Uxj2I)

So the value has to be divided by thousand.  I am not sure how to do it here in Github though...

astafan8:
@jpsecher i see that you managed to to change set_to to 0.001 - is everything working now? if so, let's merge?

PR #236, opened 2023-08-23
Add Thorlabs KLS1550 laser driver

iago-rst:
This request adds a wrapper class for Thorlabs Kinesis DLL and a driver for Thorlabs KLS1550 laser source using the DLL. 

The existing Thorlabs APT wrapper did not seem to work with KLS1550 even though it should be supported for it (only driving the laser with the APT GUI worked, but not directly sending commands through the ATP DLL).

### DIFF
--- a/docs/examples/Thorlabs_KLS1550.ipynb
+++ b/docs/examples/Thorlabs_KLS1550.ipynb
@@ -35,7 +35,7 @@
    "id": "premium-batman",
    "metadata": {},
    "source": [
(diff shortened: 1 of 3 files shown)


astafan8 (approved):

astafan8:
@iago-rst thank you for the PR! Could you disable execution of the example notebook duing docs build? here are the isntructions on how to do that https://qcodes.github.io/Qcodes/examples/writing_drivers/Creating-Instrument-Drivers.html#Documentation . Then the PR will be merged!

PR #238, opened 2023-08-24
support MontanaInstruments Cryostation

julienbarrier:
* tested on Montana Instruments Cryostation s50
* using `IPInstrument` because this instrument cannot be found as a Visa resource

### DIFF
--- a/docs/examples/Thorlabs_KLS1550.ipynb
+++ b/docs/examples/Thorlabs_KLS1550.ipynb
@@ -1,0 +1,190 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "5658297a",
+   "metadata": {},
+   "source": [
+    "# QCoDeS example with Thorlabs KLS1550 laser source\n",
+    "(same .dll as for KLS101, KLSnnn)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "1a175bbe",
+   "metadata": {},
+   "source": [
+    "## Initialisation\n",
+    "Import all required libraries for driving the devices."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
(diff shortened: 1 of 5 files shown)


qcodes_contrib_drivers/drivers/MontanaInstruments/cryostation.py
```diff
@@ -0,0 +1,182 @@
...
+from typing import Optional, Any, Dict
+
+from qcodes.instrument import IPInstrument
+from qcodes.parameters import Parameter
+from qcodes import validators as vals
+
+log = logging.getLogger(__name__)
+
+
+class MontanaInstruments_Cryostation(IPInstrument):
```
astafan8, line 21: can't this instrument be connected to visa VISA? So the driver class inherits from VisaInstrument, and then for the adress you pass something like `TCPIP[board]::host address::port::SOCKET` (see here https://pyvisa.readthedocs.io/en/latest/introduction/names.html#visa-resource-syntax-and-examples )?

astafan8: IPInstrument is not recommended to be used, unless really needed, it's an old "hack" implementation in a way, so if it's possible to use VISA that'd be much more robust in terms of connectivity as well.

astafan8 (approved):

qcodes_contrib_drivers/drivers/MontanaInstruments/cryostation.py
julienbarrier: Unfortunately this raises a VisaIOError. Trying to find the resource directly from pyvisa or NI Max is impossible.

astafan8: ok :(

PR #243, opened 2023-08-29
add SR_7270 driver with example notebook

simonzihlmann:
This PR adds a driver for the SR7270 lock-in from Ametek.
Thanks to Elyjah Kiyooka for the work!

### DIFF
--- a/qcodes_contrib_drivers/drivers/Ametek/SR_7270.py
+++ b/qcodes_contrib_drivers/drivers/Ametek/SR_7270.py
@@ -1,0 +1,296 @@
+# This Python file uses the following encoding: utf-8
+
+"""
+Created by Elyjah <elyjah.kiyooka@cea.fr>, Jan 2022
+
+"""
+
+from qcodes import VisaInstrument, validators as vals
+from qcodes.utils.delaykeyboardinterrupt import DelayedKeyboardInterrupt
+from qcodes.utils.validators import ComplexNumbers
+
+class Signalrecovery7270(VisaInstrument):
+    """
+    This is the qcodes driver for the Ametex Signal Recovery
+    Model 7270 DSP Lockin amplifier
+
+    Note:
+    ask_raw command has been rewritten (bottom) to also read an echo to remove from buffer.
+    write_raw command have been rewritten to also read after writing using ask_raw.
+
+    """
+    def __init__(self, name: str, address: str, **kwargs):
+        super().__init__(name, address, terminator='', device_clear = True, **kwargs)
+
+        idn = self.IDN.get()
+        self.model = idn['model']
+
+        self.add_parameter(name='x',
+                        label='Lock-In X',
+                        get_cmd='X.',
+                        get_parser=float,
+                        unit='V',
+                        vals = vals.Numbers(),
+                        docstring="Gets X lockin component in V; "
+                                "only gettable.")
+
+        self.add_parameter('y',
+                        label='Lock-In Y',
+                        get_cmd='Y.',
+                        get_parser=float,
+                        unit='V',
+                        vals = vals.Numbers(),
+                        docstring="Gets Y lockin component in V; "
+                                "only gettable.")
+
+        self.add_parameter('xy',
+                        label='Lock-In XY Complex',
+                        get_cmd=self._get_complex_voltage,
+                        unit='V',
+                        vals = ComplexNumbers(),
+                        docstring="Complex voltage parameter "
+                                "calculated from X, Y phase using "
+                                "Z = X + j*Y")
+
+        self.add_parameter(name='r',
+                        label='Lock-In R',
+                        get_cmd='MAG.',
+                        get_parser=float,
+                        unit='V',
+                        vals = vals.Numbers(),
+                        docstring="Gets magnitude of XY lockin components in V; "
+                                "only gettable.")
+
+        self.add_parameter(name='phase',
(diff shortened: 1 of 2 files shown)


qcodes_contrib_drivers/drivers/Ametek/SR_7270.py
```diff
@@ -0,0 +1,296 @@
+# This Python file uses the following encoding: utf-8
+
+"""
+Created by Elyjah <elyjah.kiyooka@cea.fr>, Jan 2022
+
+"""
+
+from qcodes import VisaInstrument, validators as vals
+from qcodes.utils.delaykeyboardinterrupt import DelayedKeyboardInterrupt
+from qcodes.utils.validators import ComplexNumbers
```
astafan8, line 10: please use imports from qcodes as declared as public api here https://qcodes.github.io/Qcodes/api/index.html
```suggestion
from qcodes.instruments import VisaInstrument
from qcodes import validators as vals
from qcodes.utils import DelayedKeyboardInterrupt
from qcodes.validators import ComplexNumbers
```

qcodes_contrib_drivers/drivers/Ametek/SR_7270.py
```diff
@@ -0,0 +1,296 @@
...
+    This is the qcodes driver for the Ametex Signal Recovery
+    Model 7270 DSP Lockin amplifier
+
+    Note:
+    ask_raw command has been rewritten (bottom) to also read an echo to remove from buffer.
+    write_raw command have been rewritten to also read after writing using ask_raw.
+
+    """
+    def __init__(self, name: str, address: str, **kwargs):
+        super().__init__(name, address, terminator='', device_clear = True, **kwargs)
```
astafan8, line 23: this way the terminator can be changed without changing the driver code
```suggestion
    def __init__(self, name: str, address: str, terminator='', **kwargs):
        super().__init__(name, address, terminator=terminator, device_clear = True, **kwargs)
```

qcodes_contrib_drivers/drivers/Ametek/SR_7270.py
```diff
@@ -0,0 +1,296 @@
...
+        with DelayedKeyboardInterrupt():
+            response = self.visa_handle.query(cmd)
+            if response.endswith('\x00'):
+                resp = response[:-1]
+                if resp.endswith('\n'):
+                    return resp[:-1]
+                else:
+                    return resp
+            else:
+                raise RuntimeError(response)
```
astafan8, line 264: i'd recommend to reimplement this with using ``super().ask_raw(...)`` and then add additional processing if needed. Moreover, i don't think this additional processing is needed - the pyVISA can do this for you if you set the terminator to be '\n\x00' - have you tried that?

qcodes_contrib_drivers/drivers/Ametek/SR_7270.py
simonzihlmann: thanks, good point.

qcodes_contrib_drivers/drivers/Ametek/SR_7270.py
simonzihlmann: @astafan8 I have found some time to look into this. We do the processing of the response due to the fact that the device not always responds with the same terminator. Read commands are terminated with "\n\x00" and write commands are acknowledged with a return of only "\x00".
So you say that a single terminator of "\n\x00" would still work?

Concerning the reimplementation using `super().ask_raw(...)`: this means you would replace `response = self.visa_handle.query(cmd)` with `response = super().ask_raw(cmd)`?

astafan8: > oncerning the reimplementation using super().ask_raw(...): this means you would replace response = self.visa_handle.query(cmd) with response = super().ask_raw(cmd)?

yes :)

> Read commands are terminated with "\n\x00" and write commands are acknowledged with a return of only "\x00".
So you say that a single terminator of "\n\x00" would still work?

oh, i see. No, single terminator is unlikely to work here. What you can do is to set read and write terminators individually on the visa_handle level like here: https://github.com/QCoDeS/Qcodes/blob/c596700363ff23967c10ea49d3161f614ca119b7/src/qcodes/instrument/visa.py#L234C1-L235C59 . You can implement this in the following way: in the `__init__` of the driver, when you call `super().__init__(....)`, pass `terminator=None` as the argument so that the qcodes VisaInstrument does not set the terminator at all, and then after that set the write and read terminators via `visa_handle`. It would also be a good practice to add read_terminator and write_terminator as keyword arguments to the driver that you're writing so that (just in case) the terminators can be changed by the user of the driver.

simonzihlmann: Ok, then if I understood correctly, we would not need at all the redefinition of `ask_raw `and `write_raw `functions since pyvisa would take care of the different terminators.
I will have to check this in the lab with an actual device.

simonzihlmann: I did not manage to solve this issue with the terminator. The hack checking for the terminators in the `ask_raw` function works fine and I will not dig deeper to solve this. I hope otherwise the driver is fine.

astafan8: ok

astafan8 (approved):

PR #244, opened 2023-08-30
Add Thorlab drivers for KDC101, TDC001, PM100D; extend Kinesis libraries to be more general and support more drivers

julienbarrier:
- Add drivers for:
  * KDC101
  * TDC001
  * PM100D
- refactor the Kinesis libraries to include a large range of drivers + potentially new types of devices
- place APT modules and Kinesis into a private subfolder (in the model of the Rohde&Schwartz drivers)

Tested with KDC101, TDC001, PM100D.
The KLS1550 driver was edited to allow a more general Kinesis library, but could not be tested with a real device.


qcodes_contrib_drivers/drivers/Thorlabs/PM100D.py
```diff
@@ -0,0 +1,186 @@
...
+            set_parser=lambda val: float(val)*1e3,
+            vals=vals.Numbers(),
+            instrument=self
+        )
+
+        self.write('STAT:OPER:PTR 512')
+        sleep(.2)
+        self.write('STAT:OPER:NTR 0')
+        sleep(5)
+        self.ask('STAT:OPER?')
```
astafan8, line 159: i'd recommend to create methods (or even parameters) for these 3 commands, that way the driver code becomes more readable in that it's easy to see "what is happening" rather than "which commands are being sent" which can be very cryptic

astafan8 (approved):

PR #248, opened 2023-09-04
Upgrade package infrastructure and drop 3.8 support.

jenshnielsen:
* Convert all setup from setup.cfg to pyproject.toml now that its fully supported by setuptools
* Bump min setuptools version to match
* Drop 3.8 support
* Add 3.11 as supported


astafan8 (approved):
thank you so much Jens!!

PR #261, opened 2023-10-18
Add a bunch of new drivers

thangleiter:
I am finally tackling our driver backlog. Some of these are likely to be of limited broad appeal, so let me know if adding them is out of scope for the main repo.

If desired I can split up the PR into one for each new driver.


docs/examples/Horiba_FHR.ipynb
astafan8, line 1: please add a header as the first cell of this notebook such taht it get meaningfully rendered in the docs - so e.v. ``# Example of Horiba FHR``

astafan8, line 1: if not done yet, remember to turn off execution of this notebook upon docs build as described here https://qcodes.github.io/Qcodes/examples/writing_drivers/Creating-Instrument-Drivers.html#Documentation

astafan8 (approved):

astafan8:
also a few typing errors to fix, see here https://github.com/QCoDeS/Qcodes_contrib_drivers/actions/runs/6561129954/job/17820255378?pr=261

astafan8:
thanks for the contribution!

thangleiter:
> also a few typing errors to fix, see here https://github.com/QCoDeS/Qcodes_contrib_drivers/actions/runs/6561129954/job/17820255378?pr=261

I'd appreciate if you could help me out with some of those since I don't have a lot of experience with mypy. 

- [The first one](https://github.com/QCoDeS/Qcodes_contrib_drivers/actions/runs/6561129954/job/17820255378?pr=261#step:7:15) looks to me like it's unavoidable if msl-loadlib is not added to the requirements, which I'm not sure we want.
- [This and the two following ones](https://github.com/QCoDeS/Qcodes_contrib_drivers/actions/runs/6561129954/job/17820255378?pr=261#step:7:19) seems like `ctypes.byref` doesn't accept `TypeVar`s? I can't come up with another solution than to drop the hint altogether.
- [The stubs for `tqdm` are not installed](https://github.com/QCoDeS/Qcodes_contrib_drivers/actions/runs/6561129954/job/17820255378?pr=261#step:7:31). Is there another way around this?
- [Stacking `classmethod` and `property` decorators](https://github.com/QCoDeS/Qcodes_contrib_drivers/actions/runs/6561129954/job/17820255378?pr=261#step:7:35) apparently [has been deprecated in Python](https://github.com/python/cpython/pull/92379). Any suggestions for how instead to enforce an abstract attribute that subclasses need to implement? I also used this in #262 :)

astafan8:
* `msl.loadlib` you can add to `pyproject.toml` to `tool.mypy.overrides` section here https://github.com/QCoDeS/Qcodes_contrib_drivers/blob/3c10e0759c910077c159e4f79c508dab061fb34f/pyproject.toml#L66-L78
* `types-tqdm>=4.64.6` can be added to `pyproject.toml` in `project.optional-dependencies` `[test]` target
* for the classmethod property, i found a suggestion here https://stackoverflow.com/a/76047344 for using a metaclass. So you can go that route. Or, which I encourage, is to think why it's needed - why is it important to have a class-property (not an instance-property) - perhaps a simple class-attribute would work?

astafan8:
> qcodes_contrib_drivers/drivers/M2/M2_Solstis_3.py:142:37: error: X | Y syntax for unions requires Python 3.10  [syntax]

to sovle this you could add ``from __future__ import annotations`` at the top of the module

> qcodes_contrib_drivers/drivers/Thermotek/Thermotek_T255p.py:5:1: error: Module "typing" has no attribute "ParamSpec"; maybe "_ParamSpec"?  [attr-defined]

to solve this, i don't remember, but perhaps the minimum version of typing_extensions needs to be bumped or so.

astafan8:
@thangleiter instead of https://github.com/QCoDeS/Qcodes_contrib_drivers/commit/83b844be743afbf2b410cf9d63df400be9168ee4 you could add this module to autodoc mock imports like here https://github.com/QCoDeS/Qcodes/blob/main/docs/conf.py#L439

thangleiter:
Mypy finally runs without errors locally on 3.9.

thangleiter:
... Docs should also be fixed now!

PR #302, opened 2024-02-28
Siglent (SDG) drivers and Lakeshore625 drivers improvements

hp-peti:
- initial version for Siglent SDG type waveform generators and rudimentary support for SDS type scopes
- some improvements for using the Lakeshore 625 in real-world scenarios where device state needs to be retained, and including persistent switch state support (ramp algorithm not included)


hp-peti:
merged main into branch and moved the files to the src/ tree

PR #319, opened 2024-04-25
Mini circuits modular test system

FKMalina:
Our system was recently equipped with Minicircuits modular test system (https://www.minicircuits.com/WebStore/ztm.html).
I created a new driver for the instrument, which reads out the modules installed in the system and makes an Instrument object with several InstrumentModules.
The driver was inspected by Mini-Circuits, but not reviewed.
The driver only includes the switch and amplifier modules, which are not of all available modules for the instruments. However, our device has only some of the possible modules, so I limited the driver to ones I could have tested explicitly, and the ones that are extremely similar.


astafan8 (approved):
@FKMalina awesome, thank you for contributing! Would you need help with adressing typing errors (see CI)?

FKMalina:
I fixed typing error. LMK if there is something else needed to make the driver useful and easy to use!

PR #327, opened 2024-05-07
QCodes Driver SPECS Zurich

Diddleydope:


src/qcodes_contrib_drivers/drivers/SPECS/Specs_Tramea.py
```diff
@@ -0,0 +1,1524 @@
...
+    def UserOut_ValSet48(self, Output_value: np.float32):
+        self.n.UserOut_ValSet(48, Output_value)
+
+    def UserOut_ValSet49(self, Output_value: np.float32):
+        self.n.UserOut_ValSet(49, Output_value)
+
+    def UserOut_ValSet50(self, Output_value: np.float32):
+        self.n.UserOut_ValSet(50, Output_value)
+
+    def ThreeDSwp_SwpAcqChsSet(self, channelIndexes: np.array(int)):
```
jenshnielsen, line 910: Here you should probably use NDArray from numpy typing https://numpy.org/devdocs/reference/typing.html#numpy.typing.NDArray (remember to add the import at the top of the file) 

You probably want something like NDArray[np.integer] or NDArray[np.int32] if you want a specific integer size

Diddleydope: I have changed the things you mentioned. In case you're confused as to why I have "nanonis-tramea-specs" once and "nanonis_tramea_specs" the other time, that's because the actual module name, as in in order to install it, is "nanonis-tramea-specs" and it is imported as "nanonis_tramea_specs". This is bad practice and will be changed soon.

src/qcodes_contrib_drivers/drivers/SPECS/Specs_Tramea.py
```diff
@@ -0,0 +1,1524 @@
...
+        self._socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self._socket.connect((self._address, self._port))
+
+        self.n = nanonis_tramea_specs.Nanonis(self._socket)
+        # self.n.Bias_Set(np.float32(1.2))
+
+        self.add_parameter(
+            "Output1",
+            set_cmd=self.UserOut_ValSet1,
+            get_cmd=self.Signals_ValGet24,
```
einsmein, line 38: Can we directly use `self.n` attributes instead of defining individual functions for each channel?
```python
    set_cmd=partial(self.n.UserOut_ValSet, 1)
    get_cmd=(self.n.Signals_ValGet(24, 1)[2])[0]
```

To verify, is it intended that `get_cmd=self.Signals_ValGet24` (specifically 24) for all `OutputX` parameters?

Diddleydope: I was unaware that you could write partial functions like this in python (this is my first project in python). This is clearly way better. And no, that is not intended, I must have accidentally done that. I will change this immediately. Thank you for the input!

src/qcodes_contrib_drivers/drivers/SPECS/Specs_Tramea.py
```diff
@@ -0,0 +1,750 @@
...
+                get_cmd=(lambda out=out: self.n.Signals_ValGet(out + 23, 1)[2][0])
+            )
+        for inp in range(24):
+            self.add_parameter(
+                name="{}{}".format("Input", inp+1),
+                get_cmd=(lambda inp=inp: self.n.Signals_ValGet(inp, 1)[2][0])
+            )
+
+
+    def ThreeDSwp_SwpAcqChsSet(self, channelIndexes: np.ndarray[np.integer]):
```
jenshnielsen, line 49:
```suggestion
    def ThreeDSwp_SwpAcqChsSet(self, channelIndexes: npt.NDArray[np.integer]):
```
To use the single generic argument version of nd array I think you need to use the version from numpy.typing
we usually import that like this

```
import numpy.typing as npt
```

The same fix needs to be applied in a few more places

jenshnielsen (approved):

PR #331, opened 2024-05-17
Swabian Instruments Time Tagger driver

thangleiter:
This PR adds a driver for the Swabian Intsruments Time Tagger. 

See the [module docstring](https://github.com/qutech-lab/Qcodes_contrib_drivers/blob/feature/swabian_time_tagger_driver/src/qcodes_contrib_drivers/drivers/SwabianInstruments/Swabian_Instruments_Time_Tagger.py) for design decisions.


thangleiter:
I have no idea why mypy [errors here](https://github.com/QCoDeS/Qcodes_contrib_drivers/actions/runs/9130169743/job/25106442048?pr=331#step:7:15), maybe someone can help out.

jenshnielsen:
@thangleiter Looks like this needs a bit of work on the type checking with mypy. Let me know if you run into any issues with that

thangleiter:
Thanks, I could indeed use some help with the error that mypy has in the 3.9 run. The `Callable` annotation does get a `ParamSpec` as its first argument, yet still mypy complains that

```
The first argument to Callable must be a list of types, parameter specification, or "..."  [valid-type]
```

I will get to work on the other Python versions, although this is the first time I have mypy giving different errors for different versions. Let's see if the fixes don't break the other versions.

thangleiter:
> Thanks, I could indeed use some help with the error that mypy has in the 3.9 run. The `Callable` annotation does get a `ParamSpec` as its first argument, yet still mypy complains that
> 
> ```
> The first argument to Callable must be a list of types, parameter specification, or "..."  [valid-type]
> ```

Still don't understand why it did not work the way I typed it but it should be fixed now in any case 👍

thangleiter:
Hey @jenshnielsen, this is good to go from my side. Do you have time for a review?

astafan8 (approved):

PR #333, opened 2024-05-17
Several driver bugfixes and small improvements

thangleiter:
This PR makes several bugfixes and small improvements.

### Attocube ANC350
- Adds a setter for the Attocube ANC350 position parameter.

### Horiba FHR
- Adds an `active_grating` parameter.
- Use built-in offset argument for the `width` parameter of slits.
- Make port openess a parameter.
- Give channels a label.
- Wrap requests to the dll a `qcodes.utils.DelayedKeyboardInterrupt` context.

### Aviosys IP Power 9258S
- Update the instrument class name to follow convention and deprecate the old class.
- Remove the set parser for `power_enabled`.

### Lighthouse Photonics Sprout G
- Make `enabled` a `DelegateParameter` of `output_status`.

### Msquared Solstis 3
- Improve wavelength lock acquisition logic.
- Add table wavelength tuning.
- Improve snapshot behavior

### Thermotek T255p
- Drop watchdog timeout. Contrary to the manual I have, it does not appear to be necessary.
- Do not get parameters on snapshots (takes a long).


src/qcodes_contrib_drivers/drivers/Aviosys/IP_Power_9258S.py
```diff
@@ -92,3 +95,13 @@ def __init__(self, name: str, address: str, login_name: str, login_password: str
     # get functions
     def get_idn(self):
         return {'vendor': 'Aviosys', 'model': 'IP Power 9258S'}
+
+
+class Aviosys_IP_Power_9258S(AviosysIPPower9258S):
```
jenshnielsen, line 100: I would suggest using the `@deprecated` decorator from typing_extensions to mark a class deprecated. You can see an example of how that works here https://github.com/microsoft/Qcodes/blob/main/src/qcodes/instrument_drivers/american_magnetics/AMI430.py

This will allow static type checkers to discover use of the deprecated class and mark import and use of it as striked out in vscode (and perhaps other editors)

src/qcodes_contrib_drivers/drivers/Horiba/Horiba_FHR.py
```diff
@@ -6,14 +6,13 @@
...
 import sys
-from typing import Mapping, Any, Dict
-
-from qcodes import validators
-from qcodes.instrument import (Instrument, InstrumentBase, InstrumentChannel,
-                               ChannelList)
-from typing_extensions import Literal
+from typing import Any, Dict, Mapping, cast
 
+from qcodes import DelegateParameter, Parameter, validators
```
jenshnielsen, line 11:
```suggestion
from qcodes.parameters import DelegateParameter, Parameter
from qcodes import validators
```

jenshnielsen (approved):

PR #334, opened 2024-05-18
Extend Andor iDus/DU camera driver

thangleiter:
This PR significantly extends the capabilities of the Andor iDus camera driver. A fairly large part of the SDK is now covered. 

Since the central data parameter, `ccd_data`, is dynamic in that it returns different shapes of arrays depending on the acqusition and read modes, I dropped the `spectrum` moniker. However, this also means that backwards compatibility is broken. If this is something we want to avoid, one could add a `DelegateParameter` that issues a deprecation warning when called.

## To Do
- [ ] Test hardware triggering
- [ ] `acquired_pixels` needs to be aware of downsampling that, e.g., image mode performs.
- [ ] acquisition in kinetics mode with the cosmic ray filter turned on errors. 

## Missing Features / Needs testing
- Live stream of the CCD data using the `run till abort` mode. This requires some async work, but would be nice to have since taking a single picture at a time (1 or 2d) has quite a lot of communication overhead, meaning the fastest update rate I could achieve was ~200 ms.
- Shutter modes (feature not available on our camera)
- More real-time filter modes (feature not available on our camera)


thangleiter:
I think this is ready to go for now. I'll postpone work on hardware triggers until I have more time.

Do we add a newspiece here in this repo as well?

thangleiter:
Could a maintainer please review this? 🙂

astafan8:
@thangleiter apologies for late response. There seem to be conflicts on this PR, could you resolve those and then we can review/merge?

astafan8:
> How should we deal with that? In principle, I could add spectrum back to the deprecated Andor_DU401 instrument I suppose
@thangleiter i agree, just keep the old parameter as is, you can mark it as deprecated via its docstring

PR #337, opened 2024-05-21
QphoX - CryoSwitchController

FKMalina:
This is a QCoDeS driver for a QphoX cryogenic switch controller:
https://qphox.eu/switch_controller

The driver is created by making a copy of the QphoX SDK (https://github.com/QphoX/CryoSwitchController). The license allows it, but has a specific restrictions. For that reason the driver folder contains a copy of the QphoX license and readme files indicating which files come from the SDK, and which are the QCoDeS driver wrappers. This makes the code more convoluted by necessary, but I figured that keeping things clearly separated is the best way to comply with the license.

The SDK uses the files in the driver directory to keep track of the switch state. It's not beautiful. I did my best to ensure they are not synchronized by git. It is not a beautiful solution, but I did not want to make changes in the files from the QphoX SDK.

The drivers was checked by Cristobal Ferrer <cristobal@qphox.eu> from QphoX, so he can be pinged with some questions if needed.


FKMalina:
@astafan8
I can see that the driver checks fail at the point where the QphoX SDK loads `serial` from `pyserial`. Should it be added to requirements, or would you suggest to deal with that in a different way?

FKMalina:
@astafan8
Could you suggest how I can proceed so that the new driver would be merged?

PR #338, opened 2024-05-23
modified Valon_5015 and add Vaunix LDA ethernet access

kto-ts4:
### Modified Valon_5015.py
- Previously `Valon_5015.py` does not work because of failure in `_flush()` method. 
- Which buffer to be cleared was changed. Not sure this change is correct but works well now.
### Add LDA_eth.py
- It is used for connection with Vaunix digital attenuator via ethernet.
- Based on `LDA.py`, rewriting some methods to use DLL by the vendor.

### DIFF
--- a/src/qcodes_contrib_drivers/drivers/Valon/Valon_5015.py
+++ b/src/qcodes_contrib_drivers/drivers/Valon/Valon_5015.py
@@ -194,4 +194,5 @@ class Valon5015(VisaInstrument):
         self._flush()
 
     def _flush(self):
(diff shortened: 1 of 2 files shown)


jenshnielsen:
@kto-ts4 Was there a specific reason for closing this. We are a bit behind on review but we will get to all prs eventually

kto-ts4:
@jenshnielsen I noticed that I missed to include Jupyter notebook example that is required in the readme. If it is OK, I will reopen this PR.

astafan8 (approved):

PR #350, opened 2024-08-13
Added get_idn() to QuTech Instruments.

sldesnoo-Delft:
get_idn was missing in these instruments. This resulted in many warnings in the logging.


sldesnoo-Delft:
Without the driver cannot be imported.
No change to create any unit tests. 
No change to improve code coverage.

PR #352, opened 2024-08-16
Create Rigol_DG800

BenMowbray:
Driver for Rigol DG800 waveform generator


jenshnielsen:
@BenMowbray Thanks for your contribution. However, in all 4 it seems like the files do not end in a .py could you fix that and we can have a look

PR #370, opened 2024-10-17
Bug correction in smw200a driver not properly getting options

Igcig:


Igcig:
The conditional for the options SMW-B22 and SMW-B20 option always evaluates to true because the first condition of the or statement is  'SMW-B22' which always evaluates to true

jenshnielsen (approved):

PR #388, opened 2024-11-27
Add PeakTech15xx Instrument

DCEM:


src/qcodes_contrib_drivers/drivers/PeakTech/PeakTech_15xx.py
```diff
@@ -0,0 +1,409 @@
+from typing import Any, Optional
+
+import pyvisa
+
+from qcodes.instrument.visa import VisaInstrument
+from qcodes.utils import DelayedKeyboardInterrupt
+from qcodes.utils.helpers import create_on_off_val_mapping
+from qcodes.utils.validators import Numbers
```
jenshnielsen, line 8: Please update these imports to use the official locations as documented here. https://microsoft.github.io/Qcodes/api/index.html

Specifically, validators is its own submodule, VisaInstrument should be imported directly from qcodes.instrument and create_on_off_val_mapping from qcodes.parameters

jenshnielsen:
@DCEM Looks good. Left one inline comment and the CI failures will need to be resolved before merging but otherwise should be ready.

jenshnielsen (approved):

src/qcodes_contrib_drivers/drivers/PeakTech/PeakTech_15xx.py
DCEM: I'm somewhat unsure if this is a desirable solution (done because of mypy):
        self.serial_handle: SerialInstrument = cast(SerialInstrument, self.visa_handle)

I think I implemented the other changes (including explicitly assigning a parameters as an attributes meintioned in the other PR)

If there are other issues, please let me kow.

DCEM:
@jenshnielsen can you check if anything is missing?

jenshnielsen:
@DCEM sorry for the delay not sure what happened but ci never triggered so I closed and reopended

PR #396, opened 2024-12-20
add cryomagnetics driver and examples LM500 helium level meter

thealohaprotocal:
@astafan8 and @jenshnielsen could you please have a look.


astafan8 (approved):

thealohaprotocal:
Thank you for looking into this. I am not sure I understand the problem. It seems as though the (builddocs ubuntu) check fails at every cell for the ipynb file. Even cells that were just markdown comments. Perhaps there is something wrong, in general, with the formatting. Could you please advise?

PR #397, opened 2024-12-20
add cryomagnetics driver and examples for model 4G magnet power supply

thealohaprotocal:
@astafan8 and @jenshnielsen please have a look.

### DIFF
--- a/docs/examples/Cryomagnetics/Model_4G.ipynb
+++ b/docs/examples/Cryomagnetics/Model_4G.ipynb
@@ -1,0 +1,272 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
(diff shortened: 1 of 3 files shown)


astafan8 (approved):

thealohaprotocal:
Thank you for looking into this. I am not sure I understand the problem. It seems as though the (builddocs ubuntu) check fails at every cell for the ipynb file. Even cells that were just markdown comments. Perhaps there is something wrong, in general, with the formatting. Could you please advise?

jenshnielsen:
@thealohaprotocal I think the issue was an extra `,` after `{` in the cell manually added to the notebook. JSON annoyingly does not allow trailing `,` like python and most other formats

PR #408, opened 2025-02-19
Lake Shore M81-SSM driver 

abioxinst:
Addition of Lake Shore M81-SSM driver.
Plus updated _decsvisa submodule and small updates to Proteox.py and example file.


abioxinst:
Hi, 

I see this pull request has failed the mypy check. 

When I run this check myself (using latest mypy version 1.15) I don't receive the error that is being reported here.
image

Any suggestions on how to solve this? What is different about your mypy check to mine?

jenshnielsen:
@abioxinst I think the issue is version specific. Python 3.12 changed the rules about how f strings work to add additional features one of which you are using. This causes the code to fail with 3.11 but not with newer versions

src/qcodes_contrib_drivers/drivers/Lakeshore/M81_SSM.py
```diff
@@ -0,0 +1,171 @@
...
+            count = int(self.ask("TRACe:DATA:COUNt?"))
+            print(f"\rBuffered points: {count}         ", end="", flush=True)
+    
+        time.sleep(1)
+        print('\r')
+        
+        # pull data
+        string_bytes = base64.b64decode(self.ask('TRACe:DATA:ALL?')).hex()
+
+        data_format = f"<{self.ask('TRACe:FORMat:ENCOding:B64:BFORmat?').strip('\"')}"
```
jenshnielsen, line 146: One way to work around the issue in python 3.11 is to extract this command out into a value

```suggestion
        format_val = self.ask('TRACe:FORMat:ENCOding:B64:BFORmat?').strip('\"')
        data_format = f"<{format_val}"
```

jenshnielsen:
@abioxinst Sorry for the delay. I went ahead and merged this via a different pr to fix a few tiny CI issues

Issue #393, opened 2024-12-19
Reusing Valon 5015 driver for Valon 5019


paniash:
Hi,

I was wondering if it is possible to reuse the Valon 5015 driver for the [Valon 5019](https://www.valonrf.com/5019-frequency-synthesizer-20ghz-542911.html).

The 5015 driver can connect to the instrument, so no error when running:
```python
from qcodes_contrib_drivers.drivers.Valon.Valon_5015 import Valon5015

ip = "10.0.100.28"
valon = Valon5015(name="Valon", address=f"TCPIP0::{ip}::23::SOCKET")
```

But when running `valon.status()` I get an error saying: 
```text
VisaIOError: ('VI_ERROR_INV_MASK (-1073807299): Invalid buffer mask specified.', 'getting Valon_status')
```

I'd like to know if this error is because the driver is 5015 specific only, or if the issue is something else.

astafan8:
@paniash to answer your question - could you take a look at the programming manuals of both models and see what the differences could be in the "status" command? it could also be that it's a bug or suboptimal implemntation for Valon5015, and with a small fix the driver will work for the other model. Regardless the best way is to open a pull request where the driver class for the other model is explicitly defined, and that class would also be a place to implement any differences from Valon5015.

michaelc12qe:
I have just started using a Valon 5015 and I have been seeing the same error, so it's not just the 5019. It seems to me that PR #338 contains an attempt at a bugfix, it certainly gets rid of the `VisaIOError`, but I'm seeing some weird issues with communication to the 5015 (might be unrelated tbh). I'll report back or raise a PR if I figure out what's going on.

michaelc12qe:
@paniash I raised a PR which fixes this error on my 5015, take a look if it works for you too!

PR #416, opened 2025-03-05
Valon: Fix VI_ERROR_INV_MASK bug when flushing buffer

michaelc12qe:
Also:
- Improve regex to take negative numbers when necessary
- For multiline telnet message, write once and read all lines, instead of asking multiple times and filling the buffer
- Fix typo in _set_modulation_frequency

Thanks to Colin Dietrich for help with these changes

This addresses the error in #393 for a Valon 5015, however I don't have a Valon 5019 to test with.
I think that this improves on PR #338 (the part modifying the Valon), which has a change which successfully clears the error, however it does not clear the IO buffers correctly, so if you were to successively run the `Valon5015.status` and `Valon5015.id` get commands, you would see the wrong output.

Tested on a Valon 5015 running FW 2.0a which is the latest at time of writing (https://www.valontechnology.com/5015users/5015.htm)

### DIFF
--- a/src/qcodes_contrib_drivers/drivers/Valon/Valon_5015.py
+++ b/src/qcodes_contrib_drivers/drivers/Valon/Valon_5015.py
@@ -3,6 +3,8 @@ Driver for Valon 5015 Frequency Synthesizer.
 
 Please refer to Valon's 5015 Frequency Synthesizer manual for further
 details and functionality. This model is not SCPI compliant.
+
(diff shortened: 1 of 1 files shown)


jenshnielsen:
@paniash any chance you can verify if this resolves your issue in #393

jenshnielsen (approved):
Changes looks good. Will just wait to see if it can be verified.

PR #418, opened 2025-03-13
Add driver for Windfreak SynthUSB3

EdwardLaird1:


EdwardLaird1:
I can't tell why this is failing tests on ubuntu. Can an expert please help?

jenshnielsen:
@EdwardLaird1 by default we execute notebooks in CI however in this case the notebook cannot execute since there is no instrument. Please have a look at https://microsoft.github.io/Qcodes/examples/writing_drivers/Creating-Instrument-Drivers.html#Documentation to see how to disable the execution

jenshnielsen:
@EdwardLaird1 Could you check that there is a `__init__.py` file in the Windfreak folder, this is required for the folder to be correctly included in the package. Other than this the pr should be good to go once CI is passing (se my comment above)

EdwardLaird1:
Thanks @jenshnielsen, I have made these two changes and am passing most of the tests. Is this acceptable now or do I need to pass the two Codecov tests?

docs/examples/Windfreak_SynthUSB3.ipynb
```diff
@@ -0,0 +1,235 @@
...
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "[sg_1(WindfreakSynthUSB3)] Could not connect at ASRL4\n",
+      "Traceback (most recent call last):\n",
+      "  File \"C:\\Users\\lairde\\AppData\\Local\\anaconda3\\envs\\qcodes\\Lib\\site-packages\\qcodes\\instrument\\visa.py\", line 215, in _connect_and_handle_error\n",
+      "    visa_handle, visabackend, resource_manager = self._open_resource(\n",
```
jenshnielsen, line 44: Could we remove the error output from the notebook

jenshnielsen:
We don't require testing of drivers in this repo so it's ok that the coverage drops

jenshnielsen (approved):

PR #544, opened 2026-04-08
Make several driver improvements

thangleiter:
## Attocube ANC100
- exclude `reference_position` from snapshots; getting it will move the positioners by 360-eps degrees in the worst case scenario if not previously visited.

## Horiba FHR
- add `stop()` method to stop the spectrometer from moving whatever it's currently moving
- use inherited `position` parameter over `width`

## Lighthouse Sprout G
- make power set and get follow the separation in the API (`output_power` vs `output_setpoint`)

## Qdevil QDAC2
- add set parsers for several numeric types.


src/qcodes_contrib_drivers/drivers/Horiba/Horiba_FHR.py
```diff
@@ -6,15 +6,22 @@
...
-from qcodes_contrib_drivers.drivers.Horiba.private.fhr_client import FHRClient
+from qcodes.parameters import DelegateParameter, Parameter
 from typing_extensions import Literal
 
+from qcodes_contrib_drivers.drivers.Horiba.private.fhr_client import FHRClient
+
+if sys.version_info >= (3, 13):
+    from warnings import deprecated
+else:
+    from typing_extensions import deprecated
```
jenshnielsen, line 23: For future its fine to just import this from typing_extensions. We will run pyupgrade via ruff when we drop 3.12. But its also fine to leave as is

jenshnielsen (approved):
Thanks. I think the CI issue is unrelated and due to non compatibility with qcodes 0.57.0. I will fix that

jenshnielsen:
@thangleiter Main should be passing again so if you rebase or merge main this should pass

PR #572, opened 2026-07-03
Add driver for Weinschel 8331

mgunyho:
Hi, here is a driver for the Weinschel 8331 step attenuator. This is very similar to the [qcodes built in driver for the 8320](https://microsoft.github.io/Qcodes/drivers_api/Weinschel.html#qcodes.instrument_drivers.weinschel.Weinschel8320), but it has multiple channels.

I suppose it would also be possible to modify the 8320 driver to pass the number of channels as a parameter (with a default of 1), please let me know if you want me to do that instead. (We would need to be careful with backwards compatibility, here the parameters are called attenuation1, attenuation2, etc, while the current 8320 driver has a single parameter called 'attenuation'.)

### DIFF
--- a/src/qcodes_contrib_drivers/drivers/Weinschel/Weinschel_8331.py
+++ b/src/qcodes_contrib_drivers/drivers/Weinschel/Weinschel_8331.py
@@ -1,0 +1,53 @@
+from typing import TYPE_CHECKING
+
+from qcodes.validators import Enum
+from qcodes.instrument import VisaInstrument, VisaInstrumentKWArgs
+from qcodes.parameters import Parameter
+
+
+if TYPE_CHECKING:
+    from typing_extensions import Unpack
+
+class Weinschel8331(VisaInstrument):
+    """
+    QCodes driver for Weinschel 8331 stepped attenuator.
+
+    Weinschel is formerly known as Aeroflex/Weinschel
+
+    Very similar to the Weinschel 8320 driver that comes built in with QCodes,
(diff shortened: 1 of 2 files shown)


src/qcodes_contrib_drivers/drivers/Weinschel/Weinschel_8331.py
```diff
@@ -0,0 +1,53 @@
...
+                    unit="dB",
+                    set_cmd=f"ATTN {n} " + "{0:0=2d}",
+                    get_cmd=f"ATTN? {n}",
+                    vals=Enum(*range(0, 62 + 1, 2)),
+                    instrument=self,
+                    get_parser=int,
+                    label=f"Channel {n} attenuation",
+                    docstring=f"Control the attenuation of channel {n}",
+                )
+            )
```
jenshnielsen, line 51: I suppose this is a bit of a border line case since there is only one parameter but I would normally suggest creating a channel for each attenuator.

jenshnielsen:
> I suppose it would also be possible to modify the 8320 driver to pass the number of channels as a parameter (with a default of 1), please let me know if you want me to do that instead. (We would need to be careful with backwards compatibility, here the parameters are called attenuation1, attenuation2, etc, while the current 8320 driver has a single parameter called 'attenuation'.)

I don't have strong preferences, but I think I would create a new driver that supports both instruments and uses numbered attenuators and then deprecate the old one

PR #573, opened 2026-07-03
Add driver for R&S FPL1000 series spectrum analyzers

mgunyho:
Hi, this PR adds a driver for the Rohde & Schwarz FPL1000 series spectrum analyzers. Tested on a real FPL1026 device.

I didn't check very carefully whether the parameter names are consistent with other drivers, I mostly followed naming conventions used by the manufacturer's manual / commands. Let me know if you think they should be changed and/or if you have suggestions for better names that are already commonly used in other drivers.

Note that the driver is licensed under AGPL, which hasn't been used in other drivers. I followed the example of the other existing GPL2 licensed drivers with the added LICENSE_xxx file.


jenshnielsen:
@mgunyho Thanks for your contribution. I don't think we want to go ahead and add more non MIT licenced drivers. In fact if time permits I am likely to split the GPL driver into a different packages. I think it would be better to create a package with AGPL licence. From my perspective there is no issue copying the infrastructure from this package into such a package but I am not a licence expert so there may be some issues that I am not aware of.

PR #583, opened 2026-07-31
Lake Shore M91 driver

olivia-qdox:
Adding a new instrument - Lake Shore M91 FastHall Measurement Controller

Files included:
- Driver (M91_FastHall.py) + additional file required for plotting (Contact_Check_Plots.py)
- Example script (Lakeshore_M91_FastHall_Example.ipynb)
- Simulated instrument file (Lakeshore_M91_FastHall.yaml)
- Test file for driver (test_Lakeshore_M91_FastHall.py)


src/qcodes_contrib_drivers/drivers/Lakeshore/M91_FastHall.py
```diff
@@ -0,0 +1,1070 @@
+import unicodedata
+import os
+import time
+import matplotlib.pyplot as plt
+import json
+import re
+import qcodes as qc
+from enum import auto
+from strenum import StrEnum
```
jenshnielsen, line 9:
```suggestion
from enum import StrEnum
```

StrEnum is part of the std library in all python versions that we support

src/qcodes_contrib_drivers/drivers/Lakeshore/M91_FastHall.py
```diff
@@ -0,0 +1,1070 @@
...
+from qcodes.instrument import VisaInstrument, InstrumentModule
+from qcodes import Parameter
```
jenshnielsen, line 11:
```suggestion
from qcodes.parameters import Parameter
```

jenshnielsen:
Thanks @olivia-qdox 

There are some imports that looks like they break CI

in the example we have from M91_FastHall import M91_FastHall this should import from the full part

I highlighted a few more inline. Could you address these and I will review

olivia-qdox:
Thanks, I've updated the imports, let me know if there are any other problems

olivia-qdox:
Sorry I've just realised that StrEnum doesn't seem to be working in the driver now, I will update when I have sorted out the issue

